PR #8, opened 2021-11-18
Fixed image, title and content

milexm:


chrispsommers (commented):
It's good to see the original source image file (Visio) file being added to the repo. I assume the file is edited in Visio and exported as a .png for inclusion in the .md files.

I would suggest that going forward, for new images, we strive to use the .svg file format, which has a few advantages: (1) Anyone can edit w/o a license, e.g. using draw.io desktop or similar. (2) It can be included in the .md file directly as .svg, saving a step  (it can be viewed in the .md file while editing (e.g. using VScode preview) (4) no separate .png file is needed, saving disk space. I've been taking this approach in the test HLD. Thanks.

milexm:
Hello Chris, I did change the image type to SVG. Can you see the related PR?
I am supposed to join the team as a tech writer in a couple of weeks.
Right now I do not have access to the repo just experimenting, using the
forked version.
Thanks,
Michael

chrispsommers:
Hi Michael,
I do appreciate you exploring my suggestions. I saw the changes you made. I can view the changed file in your fork and it looks fine. However, I cloned your fork and tried to open the .svg file from within Draw.io desktop and it gives me an error saying “Not a media file.” I am not an expert in .svg format, but I wonder if it is not an editable SVG file?

The true value of using .svg as the single source for diagrams is it’s entirely self-contained, can be viewed and edited using many free tools. If you still need Visio to produce the original content and export to .svg, you get some advantage (size)  in linking the .svg file in the .md file (as opposed to .png), but not all advantages. So, ideally, you’d replace the Visio file entirely with an editable .svg, if that is feasible. Worst-case, if Visio can’t export an .svg which can be edited by say Draw.io, you have two choices. Leave things as-is, or redraw the diagrams in editable .svg. It’s really a judgement call. It might be better to just use your time to produce new drawings this way and leave things as-is.

I don't mean to be pedantic about this, I'd just advise going forward we keep these things in mind. There are a lot of open-source project with complex .png files embedded in them, whose original source is probably on someone's hard drive, never to be found.

documents/SDN-Features-Packet-Transforms.md
```diff
@@ -1,26 +1,37 @@
...
-| Flow Scale | •	1+ million flows per v-port (aka ENI) |
-|  | •	50 million per DPU/Card |
-|  | o	single encap IPv4 overlay and IPV6 underlay |
-|  | o	single encap IPv6 overlay and IPV6 underlay. (This can be lower) |
-|  | o	single encap IPV4 |
-|  | o	Encap IPv6 and IPV4 |
-|  | *These are complex flows, details are below*
+| ---------- | ----------- |
+| Flow Scale <img style="width:200px"/>| <ul><li>1+ million flows per v-port (aka ENI)</li> <li>50 million per DPU/Card</li> <li>single encap IPv4 overlay and IPV6 underlay</li> <li>single encap IPv6 overlay and IPV6 underlay. (This can be lower)</li> <li>single encap IPV4</li> <li>Encap IPv6 and IPV4</li></ul>|
+|  | *Complex flows details are listed below*
```
lguohan, line 34: better to keep as markdown format.

KrisNey-MSFT:
Obsolete

PR #21, opened 2021-12-15
PNA compatible connection tracking

mariobaldi:
Compiling successfully

### DIFF
--- a/sirius-pipeline/sirius_conntrack.p4
+++ b/sirius-pipeline/sirius_conntrack.p4
@@ -3,6 +3,140 @@
 
 #include "sirius_headers.p4"
 
+#ifdef PNA_CONNTRACK
+
+#include "pna.p4"
+
+const ExpireTimeProfileId_t EXPIRE_TIME_PROFILE_NOW    = (ExpireTimeProfileId_t) 0;
+const ExpireTimeProfileId_t EXPIRE_TIME_PROFILE_LONG   = (ExpireTimeProfileId_t) 2;
+
+IPv4Address directionNeutralAddr (
+    in direction_t direction,
+    in IPv4Address outbound_address,
+    in IPv4Address inbound_address)
+{
+    if (direction == direction_t.OUTBOUND) {
+        return outbound_address;
+    } else {
+        return inbound_address;
+    }
+}
+
+bit<16> directionNeutralPort (
+    in direction_t direction,
+    in bit<16> outbound_port,
+    in bit<16> inbound_port)
+{
+    if (direction == direction_t.OUTBOUND) {
+        return outbound_port;
+    } else {
+        return inbound_port;
+    }
+}
+
+
+control ConntrackIn(inout headers_t hdr,
+                inout metadata_t meta)
+{
+  action conntrackIn_allow () {
+  /* Invalidate entry based on TCP flags */
+          if (hdr.tcp.flags & 0x101 /* FIN/RST */) {
+            set_entry_expire_time(EXPIRE_TIME_PROFILE_NOW); // New PNA extern
+            /* set entry to be purged */
+          }
+          restart_expire_timer(); // reset expiration timer for entry
+          meta.conntrack_data.allow_in = true;
+  }
+
+  action conntrackIn_miss() {
+          if (hdr.tcp.flags == 0x2 /* SYN */) {
+            if (meta.direction == direction_t.OUTBOUND) {
+               add_entry("conntrackIn_allow"); // New PNA Extern
+               //adding failiure to be eventually handled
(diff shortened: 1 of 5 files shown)


sirius-pipeline/sirius_conntrack.p4
```diff
@@ -3,6 +3,140 @@
...
+
+      idle_timeout_with_auto_delete = true; // New PNA property
+      const default_action = conntrackOut_miss; //New PNA property
+  }
+  apply {conntrackOut.apply();}
+}
+
+#endif // PNA_CONNTRACK
+
+
```
jfingerh, line 139: I do not know whether this code matches the desired data plane behavior of the DASH features, but the code Mario has written here does appear to be compatible with the PNA specification in at least most ways.

The top level sirius_pipeline.p4 file still include the v1model.p4 include file used for the v1model P4 architecture.  There is no urgent need to remove that with this PR, but I suppose the idea will be to remove that if and when the rest of the code is updated for the PNA architecture, too?

Another basic question of mine is that I see two different P4 table definitions here: conntrackIn and conntrackOut.  Is the idea that these will be two independent add-on-miss tables, each of which is only applied for one direction of traffic, i.e. one is applied only for inbound traffic, and the other only for outbound?  Again, I do not know the desired DASH data plane features well enough to say whether there should be one conntrack table applied for packets in both directions, or two separate tables, but wanted to ask to see if there was some other English description of the desired functionality that I could compare this with.

mariobaldi: I am not sure either if this code matches the original behavior. It would be beneficial if the original authors could take a look. 
I have used two tables to reproduce the same behavior of the two state graphs in the original code. My understanding is that both tables (like the state graphs in the original code) are applied in each direction and they are used to avoid using the same table twice in the same pipeline (i.e., in two separate stages). Specifically, the sequence of operations should be the following: 
1) look up a flow to check whether it is in the flow table
2) if not found, check the ACLs
3) if flow is allowed, add flow to the flow table 
The existing code uses one table (say TableIn) in step 1 and the other table (say TableOut) in step 3.
On the returning packets, the tables are used in the opposite order: the flow is found in TableOut, the ACL is skipped and the flow is added in TableIn as well.
Thereafter, lookups in both tables result in a hit. 
Again, this is my understanding and it would be important that it were verified by the authors of the original code.

Agreed with v1model.p4
The code also still contains the list match kind.

KrisNey-MSFT: @marian-pritsak and DASH Community, may we review / comment WRT this PR please?

sirius-pipeline/sirius_outbound.p4
```diff
@@ -91,18 +92,26 @@ control outbound(inout headers_t hdr,
...
+           ConntrackOut.apply(0);
 #endif /* STATEFUL_P4 */
 
+#ifdef PNA_CONNTRACK
+        ConntrackOut.apply(hdr, meta);
+#endif // PNA_CONNTRACK
+
         /* ACL */
-        if (meta.conntrack_data.allow_out) {
+        if (!meta.conntrack_data.allow_out) {
```
lguohan, line 103: is this a bug in current version? @marian-pritsak

mitalum:
```
if (hdr.tcp.flags & 0x101 /* FIN/RST */) {
            set_entry_expire_time(EXPIRE_TIME_PROFILE_NOW); // New PNA extern
            /* set entry to be purged */
}
```
It seems that the behavior for a FIN is to remove the connection immediately. Is this really the desired behavior, since the endpoints may still send ACKs after the FINs?

mariobaldi:
Hi John,

 

You have a very good point.

 

In the PNA compliant code I have implemented exactly the same logic as the code based on state graphs that is checked into the repository. 

My understanding is that, in general, the current P4 code is providing a rough model that is deliberately ignoring several details and corner cases. 

 

Thanks,

---

Mario

sirius-pipeline/sirius_outbound.p4
KrisNey-MSFT: Per Marian, it is a bug and will review full PR by 2/15

sirius-pipeline/sirius_conntrack.p4
venkat-pullela-keys: Theoretically, we can use a single table. If we have to guess from the code, the reason for the symmetric key generation is to use a single table.

marian-pritsak:
Hi Mario,

Can you please adapt this PR to the current directory structure so we can merge it?

mmiele:
@mariobaldi I merged your changes. Please, verify.

PR #41, opened 2022-01-25
Updated diagram, HLD text.

chrispsommers:


chrispsommers:
Please see updated "layers" diagram, it needs a label in one box.

mmiele:
Updated diagram as we discussed.

PR #56, opened 2022-02-16
Kick off HA protocol discussion with an initial proposal

budgrise:


chrispsommers:
Bud, thanks for submitting this! I'm glad we restructured the repo just this week. Can I make a couple of suggestions (which don't bear on the content itself)? I'll respond to the
content later.

- perhaps name the document "xsight-labs-ha-proposal-v1.md" or something similar, to distinguish it from the existing "HA" document in that same directory? Use -v1 in case new proposals come from your company. (I know we can use git version tracking for changes, but if you ever produce a significantly different version, it might become -"v2" etc.
- Add a link to it in the ToC table in https://github.com/budgrise/DASH/blob/main/documentation/high-avail/design/README.md

chrispsommers (approved):
Hi, I like this PR, it's very clear and concise and well-thought out. I don't have a lot of feedback on the algorithms etc., they seem logical and sound. Here are some other thoughts:
- I'd lean towards gRPC for the control channel, it's the favored messaging protocol these days and would make it easy to write production as well as test applications. It can actually be quite performant, especially if streaming connections are used.
- We need to think about a "management" interface to the HA subsystem, to configure operating parameters (e.g. timeouts), perform resets, query states and statistics, etc. This then begs the question, "will there be a SAI API to the HA controller?"
- Presumably a behavioral model could be built which allows the algorithm to be evaluated and simulated at scale, without requiring actual HW targets. (Extra credit: this could even be plugged into e.g. NS3 simulations).
- We need to think about test & performance criteria, even test techniques. If we pose some requirements, we need to ensure they're testable. This might require extra observability  features (streaming internal state & stats updates, above and beyond the presumed management API).
- We need to think about continuous observability, including a high-fidelity mode (e.g. streaming high-speed updates of the internal states) to be able to judge performance of the HA system in a lab or even live environment.

documentation/high-avail/design/ha-design.md
```diff
@@ -0,0 +1,140 @@
...
+This document is a placeholder for the HA protocol proposal.  
+The purpose is to generate discussion that will lead to an actual design doc.
+
+
+# HA Requirements
+
+The High Availability document lists most of the requirements:
+https://github.com/Azure/DASH/blob/main/documentation/high-avail/design/high-availability-and-scale.md
+
+Microsoft has provided some additional requirements:
```
chrispsommers, line 13: Suggest these be moved to initial AH document per https://github.com/Azure/DASH/issues/57

Issue #61, opened 2022-02-23
Detail questions on dash-handling-fragmented-packets.md


jfingerh:
These statements are made in the latest version of that file as of 2022-Feb-23:

+ If a subsequent packet arrives that is the start of a fragmented packet, the Frag ID must be used to create a new temporal flow that can be uniquely identified by the (Frag ID, DST, SRC) tuple.

+ If the connection is closed with the arrival of the FIN packet then all temporal flows must be closed as well.

The last statement seems to be assuming that temporal flows can be associated with a 5-tuple.  That is true if the first IP fragment arriving for one original unfragmented packet contains the L4 header.

Is there a preferred behavior if the first IP fragment for a particular (Frag ID, DST, SRC) tuple is a non-first fragment, and contains no L4 header information, and thus cannot be associated with a particular 5-tuple (at least not yet)?

jfingerh:
Another question that might be useful to answer in that same document: Is there some maximum IP fragmented packets per second maximum requirement for a DASH compliant product, that might be much much lower than "line rate" for the product?  e.g. something like "if a DASH device received more than 50K IP fragmented packets per second, it may discard all of those exceeding that rate".

If you want DASH compliant products to handle IP fragmented packets at line rate indefinitely, it would be good to state that explicitly.

jfingerh:
AI for @jfingerh to create a PR for the dash-handling-fragmented-packets.md document that makes this as explicit as possible.

PR #64, opened 2022-02-24
Minor clarification on handling of fragmented packets

jafingerhut:


jafingerhut:
Created in response to discussion on this issue: https://github.com/Azure/DASH/issues/61

chrispsommers:
I think https://github.com/Azure/DASH/pull/63/ and https://github.com/Azure/DASH/pull/64 are both trying to address the same issue, somewhat differently.

jfingerh:
Everything in PR #63 looks reasonable to me, but I do not see how #63 addresses the issue that non-first fragments must be dropped if there is no matching temporal flow state that matches it.

chrispsommers:
It seems like PR64 is a superset. Perhaps the two authors @jafingerhut and @mhanif can agree on a single unified PR? Thanks, it's nice to have a surplus of enthusiasm!

jfingerh:
If I understood correctly, in the meeting today Gerald suggested I create a PR for Microsoft engineers to look at and think about when they are doing their careful review of IP fragment handling, and likely this PR will be closed without being merged, as they will make their own changes.  It is fine with me either way.

KrisNey-MSFT:
Hi Andy, I merged and closed PR 63, would you mind resyncing and submit 64 again so I can view the changes?
Thank you very much, Kristina

jfingerh:
@KrisNey-MSFT I have merged the latest changes in the main branch with the changes proposed in this PR.

PR #67, opened 2022-02-26
doc-routing-guidelines 

mmiele:
Started article.

### DIFF
--- a/documentation/general/design/dash-routing-guidelines.md
+++ b/documentation/general/design/dash-routing-guidelines.md
@@ -1,0 +1,8 @@
+---
+title: Routing guidelines
+last update: 02/25/2022
+---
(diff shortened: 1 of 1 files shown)


chrispsommers:
IMHO it seems like a "dataplane" topic to me, just like the IP fragments topic.

mmiele:
@chrispsommers Moved to dataplane/design.

chrispsommers:
Hi @mmiele I've been reading this as it develops, it looks very helpful and well-written. However, should the use of the term "VM" be replaced with "DASH DPU?" I might be uninformed, but it seems like this is the "legacy" architecture and DASH DPUs are meant to accelerate what was formerly done in a VM.

PR #72, opened 2022-03-03
sdn-features-packet-transforms-images 

mmiele:
Replaced 2 images.


KrisNey-MSFT (changes requested):
Looks like we need to remove the piece of the image that says 'SDN' and the 'GS Control Plane' text.

mmiele:
@KrisNey-MSFT I fixed the images. Please, verify.

KrisNey-MSFT (approved):
Looks good to me, thank you

PR #77, opened 2022-03-10
Connection Tracking - Flow Creation and Termination

anjalisinghai1:
**Connection Tracking flow creation and termination using Absolute timers, that do not get replenished when next set of packets come in for the flow.**

A new timer is used for flow creation (state 1) and another one for flow termination(state 3) which is absolute in nature.
A different replenishing timer is used during Established state (state 2) of the flow.

It is important to configure the absolute timers to a small enough value to minimize percentage of stale flows in the flow cache and also to not affect the throughput by making it too small.

**Flow: (Asymmetric Firewall usage)**
        Note: Symmetric firewall is a special case of the asymmetric one.

1. When a flow starts from a trusted direction (if this direction is allowed to Auto Insert CT rules in the Stateful Table) it will first be checked in the Stateless Firewall Policy table.
2. If there is a match in the stateless firewall Policy table and the packet meets certain criteria such as presence of SYN flag then an Auto Insert action is generated by the Stateless Table for the Stateful table.
3. The packet goes through the Stateful table and if there isn't a flow entry already in the Stateful table corresponding to this flow, a new flow entry is added for a limited time (Absolute timer 1, state =1)
4. If a packet comes from the untrusted direction at this point, it will not find a rule in the Stateless Firewall table to permit the flow since its from untrusted direction, although the Packet will not be dropped at this point.
5. The packet will flow through the Stateful firewall table and if there is a pre-install flow entry that matches this packet, then this packet is allowed to be forwarded to the destination selected by the forwarding table, otherwise the packet is dropped.

**State Transition Details:**

**State = New:**
The first SYN + NO ACK will do TIME_SEL_VAL = 1, and select the Timer to be one out of the four supported (say TIME_SELECT_TIMER_1) using Wild Card Match  for SYN flag only. In addition, the first SYN will do an add on miss if it matched on Stateless Firewall policies from previous blocks (In some cases this stateless firewall check may not be needed, and the stateless firewall policy could decide that any connection starting from a trusted source (typically a Host vport) is always allowed to add on Miss).
This will also generate a Mirror towards the Control plane.
When the SYN+ACK (first or second) comes, Wild Card Match rule will do an action of TIME_SEL_VAL = 0 or Do not change timer explicitly. If a mirror to control plane is required it can set a mirror action. This will cause CT flow rule to continue to age out based on TIME_SELECT_TIMER_1 because it was not replenished.
If Control plane is keeping track of the state of the flow, it should mark the state of this flow as established after the final SYN+ACK mirror comes.
If the connection does not get established in TIME_SELECT_TIMER_1 duration, the connection will timeout and the flow rule will be removed.

**State = Established**
WCM will match on SYN + FYN+ RST flags being CLEAR, action will do TIME_SEL_VAL = 1, TIME_SELECT_TIMER_2
This will reload the time with TIME_SELECT_TIMER2 value and the timer will start counting down.
The timer will get replenished with TIME_SELECT_TIMER2 with every new packet coming in that matches the flow.
The entry will be deleted only after packets stop matching the rule and the timer times out because it was not replenished on arrival of a packet.

**State = Tear Down using FIN**
The first FIN + NO ACK will do TIME_SEL_VAL = 1, and select  TIME_SELECT_TIMER_3 using WCM match for FIN flag only. The timer will start aging from here on as it is an absolute timer.
This can also generate a Mirror towards the Control plane if needed. 
When the FIN+ACK comes, WCM rule will do TIME_SEL_VAL = 0 or Do not change timer explicitly or do not add a rule in WCM for FIN+ACK which will also result in no Time replenish whatsoever. Flow rule will continue to age based on TIME_SELECT_TIMER_3

**State = Tear Down using RST**

RST packet will do TIME_SEL_VAL = 1, and select the timer to be TIME_SELECT_TIMER_3 using WCM match for RST flag only. The timer will start aging from here on as it is an absolute timer.
This can also generate a Mirror towards the Control plane if needed. 

**State = Established and a SYN or SYN+ACK packet comes in (Malicious SYN flood attack)**

This packet will generate mirror towards control plane, as described in State= New. Since The control plane has already transitioned to Established state for this flow, it could detect this as SYN flood attack and it can update the flow rule to drop action so as to not allow this flow to proceed any more.

**Immediate Rule removal from SW or HW**
Any Event in HW or SW can be programmed to cause the immediate aging of the rule by making TIME_SELECT_VAL = 1 and timer set to TIME_SELECT_TIMER_4 (Value = 0)

To Be added: Flow for Immediate flow removal for Second FYN-ACK


reshmaintel:
@KrisNey-MSFT @GeraldatMicrosoft Hi Kristina and Gerald, Here is the PR for Flow creation/termination handling, please take a look. Thanks.

KrisNey-MSFT:
Thank you very much!

prsunny:
@KrisNey-MSFT , @reshmaintel , closing this PR as it is raised against a private branch. Please raise a new PR to main branch if this is still pursued.

PR #92, opened 2022-04-13
Add HA Diagrams

chrispsommers:
Replaced PNG file of original HA architecture with editable SVG replica. Added a proposed SVG diagram of the Smart Switch HA configuration based on what I know, but probably needs corrections or refinements. This is intended to support upcoming HA WG discussions.

### DIFF
--- a/documentation/high-avail/design/high-availability-and-scale.md
+++ b/documentation/high-avail/design/high-availability-and-scale.md
@@ -53,9 +53,11 @@ Protocol Requirements:
 ## Architecture
 
 ### High-level architecture of Rack Design for High Availability and Scale
-Two architectures are presented.
-* The initial, near-term solution uses separate "T0" or "ToR" switches wired to "DASH Appliances" carrying six, 200GB (2x100GB ports each) DPUs. DPU redundancy requires each DPU to be wired to both switches (one 100GB link to each) as shown.
-* The second, longer-term solution uses integrated Smart Switches, containing captive DPUs wired to a main switching ASIC(s); DPU ports are not wired to front panels. This also means the DPU count and physical port configuration is not stipulated; only the aggregate throughput to the main switch ASIC is important. DPU redundancy requires dedicated inter-switch "cross-links."
+Two architectures are presented. In both cases, communications channels between DPUs are routed through the network.
+
+* The near-term solution uses separate switches wired to "DASH Appliances" carrying six, 2-port DPUs. For redundancy, each DPU is wired to both switches (one link to each) as shown.
+* The longer-term solution uses integrated Smart Switches, containing captive DPUs wired to a main switching ASIC(s); DPU ports are not wired to front panels.
+
 
 See the following diagrams.
 


GeraldatMicrosoft (approved):
The drawing that shows the cross-links should be updated. The cross link should show a network between as they are not necessarily physically connected. The comment should say communications is via the network of two co-located or remotely located nodes. The switch should not indicated ToR A ToR B or ToR C. These are just SmartSwitches that are used as appliances and can be located anywhere in the network. I would just call them DASH Appliance/SmartSwitch. Also there should be no mention of Tier2. The switches above the DASH Appliance/SmartSwitch should just be referred to as a Leaf/Spine. The 1.6Tbps should be replaced with N x Links. The Main ASIC should say Switch ASIC. After making these changes I am ok for this to be placed into the HA document.

The same goes for the first diagram. No mention of ToR, eliminate numbers and just say N x Uplinks per DASH SmartSwitch, Inter-DPU HA sync messages are routed through Leaf/Spine layer without mention of 1.6Tbps total and finally the to T2 switches should just say to higher Spine Switching layer. Drop any mention of 100GB links and use the term 100 Gbps when required as GB could be understood as GigaByte. Remove the T0 switches comment off the side and just say SmartSwitches. Where the Cross-wire Active-Active 6 DPUs per appliance is labeled just say Cross networked active active links without any mention of 1.2TB. Speeds/Feeds and Capacity are very specific so remove all occurrences and create more general labeling.

chrispsommers:
@GeraldatMicrosoft Thanks for the very specific feedback, it's just what I needed!

GeraldatMicrosoft:
great

Sent from my iPad

On May 3, 2022, at 2:13 PM, Chris Sommers ***@***.***> wrote:

﻿

@GeraldatMicrosoft<https://nam12.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FGeraldatMicrosoft&data=05%7C01%7C%7Cab4cfec80c484c9cf03d08da2d49b9b4%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C637872091918840447%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=Qm3VtiLKPIqqOREVc6DXlg%2BXaNqJKwvb937lSeJmENI%3D&reserved=0> Thanks for the very specific feedback, it's just what I needed!

chrispsommers:
Hi @GeraldatMicrosoft please take a look at the updated diagrams, I've tried to incorporate your feedback. Thanks.

GeraldatMicrosoft (commented):
In top diagram we should just say “to upper Tiered switches or network”
In lower diagram we should also say the same thing
Also in lower diagram the cross-links are logical and not necessarily or even typically physical
The communications is just through the networks so we should not imply otherwise … make the cross-link dotted and just say “logical-link”
On the bottom where you show Switch C connects to HA … again just say “through switched network” … but it is not necessary as the two appliances do not have to be in the same rack/row or even in the same datacenter (although we try and co-locate within a datacenter for latency reasons)

mmiele (approved):

PR #93, opened 2022-04-13
doc-json-config-example

mmiele:
This is WIP to start a discussion on this topic and gather relevant information and feedback.


chrispsommers (changes requested):
It's not clear to me where this schema fits in. I get that it represents a bunch of DASH service configs, but...

Does it relate to the gNMI northbound API schema? In other words, is it a faithful representation of the expected gNMI config objects?

Is it merely an informal conceptual sketch of eventual config objects seen on the gNMI northbound?

Or does it relate to some other domain, e.g. inside an SDN controller's DB?

Alternatively, is it a JSON representation of [DASH_APP_DB](https://github.com/Azure/DASH/blob/main/documentation/general/design/dash-sonic-hld.md#32-dash-app-db) objects? In this case, shouldn't the JSON schema be derived from [sonic-cfggen](https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-config-engine/sonic-cfggen)? 

IMO there should be a definitive (JSON and/or YAML) schema which is 1:1 tied to an existing schema, e.g. gNMI YANG <-> JSON or CONFIG/APP_DB <-> JSON. We should avoid yet another schema which is unrelated to anything else.

I've created https://github.com/Azure/DASH/pull/94 with a figure to more clearly show what I mean.

mmiele:
@chrispsommers I think @KrisNey-MSFT, @prsunny and @mzms could address some of your questions. Obviously, any suggestions from the community are welcome. Really, to the best of my knowledge, one of the intents of this PR (document) is to have a discussion on this important matter and come to an agreement. Obviously, you are already doing it, and thank you for that.

prsunny (approved):

prsunny:
> It's not clear to me where this schema fits in. I get that it represents a bunch of DASH service configs, but...
> 
> Does it relate to the gNMI northbound API schema? In other words, is it a faithful representation of the expected gNMI config objects?
> 
> Is it merely an informal conceptual sketch of eventual config objects seen on the gNMI northbound?
> 
> Or does it relate to some other domain, e.g. inside an SDN controller's DB?
> 
> Alternatively, is it a JSON representation of [DASH_APP_DB](https://github.com/Azure/DASH/blob/main/documentation/general/design/dash-sonic-hld.md#32-dash-app-db) objects? In this case, shouldn't the JSON schema be derived from [sonic-cfggen](https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-config-engine/sonic-cfggen)?
> 
> IMO there should be a definitive (JSON and/or YAML) schema which is 1:1 tied to an existing schema, e.g. gNMI YANG <-> JSON or CONFIG/APP_DB <-> JSON. We should avoid yet another schema which is unrelated to anything else.
> 
> I've created #94 with a figure to more clearly show what I mean.

@chrispsommers , this is a reference model and forms the basis for generating the DASH APP_DB schema. However this is not a complete one-one representation and we have made many modifications to be aligned with Sonic model for the schema.

chrispsommers:
@prsunny Thanks for responding. If I understand you, it will eventually correspond exactly to the schema as sonic-cfggen would use for import-export? Was the diagram in https://github.com/Azure/DASH/pull/94 useful? Thanks.

lisahnguyen:
I have a few questions for the examples. I am not able to add them to the corresponding of the document. Hence, I will list them all here

1. For the per-ENI’s “qos_settings”, what is the expected behavior when the limit is reached? For example, 
- When CPS rate is reached, is it expected that new connections won’t be learned even though the overall CPS per DPU has not maxed out? 
- When CPS or max flows are reached, should a per ENI counter be incremented to track the number of dropped flows due to max limit? Should all or a subset of the dropped flows be sent to the Control Plane/Controller?

2. There is a reference to per DPU initial configuration in the routing action comment “list of active modules impacts memory and needs to be configured as part of initial card config”. What other per DPU initial configurations should be expected (e.g. global connection aging time, etc.)? 

3. For each routing type, please help list a set of valid actions associated with it and any expected action orders in addition to the max number of actions. This info is helpful in writing test cases to validate expected behavior not explicitly listed in the sample json file. 

4. In term of routing action combinations, Is “maprouting” action the only exception or it is just one example and there are more exception? 

5. For the “drop” routing-type, what are the other two action types that can be associated with “drop” action-type? 

6. Please help add more info on what objects can be modified and the expected behaviors. For example, what is the expected behavior for existing connections when  ACL rules change from “allow” to “deny” that impacts those connections?

KrisNey-MSFT:
Hi @Lisa ***@***.***>, I added you as a contributor to the Repo - I believe you'll need to Accept the invitation...

***@***.***

Thank you, Kristina

lisa:
# `@lisa` User Notification

Hello! You're receiving this notice because you've tagged me, @lisa, on GitHub, likely in error. Lisa is a common first name, and happens to be mine. I use the @lisa username here on GitHub. While being tagged in various repositories is an interesting way for me to discover new projects, it's likely not your intention. Did you mean to tag someone else? Did you mean to use `@lisa` instead, in a [word as word](https://en.wikipedia.org/wiki/Use%E2%80%93mention_distinction) usage?

# Next Steps

I'm unsubscribing from this issue, pull request or project so if you meant to tag someone else to notify them, I would recommend that. If you meant to use my user ID as a plain string, without having GitHub notify me, you can do so by wrapping with backticks (the \` character), like so: `@lisa`.

mmiele:
@lisahnguyen I fixed some issues related to the JSON example. I added the actual file _DASH.json_ into the PR. Now you can review  it directly. Would you mind to add your comments into the file itself. I appreciate your help. Thanks

documentation/gnmi/design/DASH.json
```diff
@@ -0,0 +1,724 @@
+{
+"qos-settings": [
```
lisahnguyen, line 2: What is the expected behavior when the CPS or max flow limit is reached? For example,

- When CPS rate is reached, is it expected that new connections won’t be learned even though the overall CPS per DPU has not maxed out?
- When CPS or max flows are reached, should a per ENI counter be incremented to track the number of dropped flows due to max limit? Should all or a subset of the dropped flows be sent to the Control Plane/Controller?

documentation/gnmi/design/DASH.json
```diff
@@ -0,0 +1,724 @@
...
+                    "vni-key": 100
+                }
+            ]
+        }
+    }
+],
+
+//
+// active routing types for mappings (active routing types modules)
+// list of active modules impacts memory area reserved for mappings and needs to be configured as part of initial card config
```
lisahnguyen, line 265: What other per DPU initial configurations should be expected (e.g. global connection aging time, etc.)?

documentation/gnmi/design/DASH.json
```diff
@@ -0,0 +1,724 @@
...
+// Order of actions: 16-bit value that is split into 4x4 bits.
+// Example: 4to6 -> staticencap -> appliance -> none (will be represented as: 0x5340)
+// When packet hits the pipeline, it looks up LPM prefix, once the prefix is found, it gets routing type and persist arguments from LPM route
+// Then next stage takes routing type and looksup order ex. 0x5340 and persists all default arguments for stages
+// Then we go 4-times thru routing actions and apply applicable action to the packet based on carried arguments.
+// Important: "maprouting" has special implementation. If it is present (0x1000), then it will be only 1 action.
+// This action first does lookup to find the routing-type again based on prefix, and then performs the routing stages.
+//
+
+"routing-types-config": [
```
lisahnguyen, line 161: For each routing type, please help list a set of valid actions associated with it and any expected action orders in addition to the max number of actions. This info is helpful in writing test cases to validate expected behavior not explicitly listed in the sample json file.

documentation/gnmi/design/DASH.json
```diff
@@ -0,0 +1,724 @@
...
+                }
+            ]
+        }
+    },
+    {
+        "vpc": {
+            "routing-type": "vpc",
+            "actions": [ // ONLY 1 action here! More then 1 action is not allowed if action-type used is "maprouting"
+                {
+                    "action-type": "maprouting" // NOTE: "maprouting" will be the ONLY action. It cannot be used with other actions
```
lisahnguyen, line 177: In term of routing action combinations, Is “maprouting” action the only exception or it is just one example and there are more exception? Please help elaborate if there are more exceptions

documentation/gnmi/design/DASH.json
```diff
@@ -0,0 +1,724 @@
...
+                    "action-type": "staticencap",
+                    "encap-type": "vxlan"
+                }
+            ]
+        }
+    },
+    {
+        "drop": {
+            "routing-type": "drop",
+            "actions": [ // up to 3 actions
```
lisahnguyen, line 196: For routing-type "drop", what are the other two action types that can be associated with “drop” action-type?

documentation/gnmi/design/DASH.json
```diff
@@ -0,0 +1,724 @@
+{
```
lisahnguyen, line 1: Please help add more info on what objects can be modified and the expected behaviors. For ACL specifically, what is the expected behavior for existing connections when ACL rules change from “allow” to “deny” that impacts those connections?

lisahnguyen:
@mmiele Thanks for providing the DASH.json file. I have moved my comment to the appropriate section as you suggested

mmiele:
@lisahnguyen Great feedback by distributing your comments in the JSON file.

mmiele:
> 2\. n



> I have a few questions for the examples. I am not able to add them to the corresponding of the document. Hence, I will list them all here
> 
> 1. For the per-ENI’s “qos_settings”, what is the expected behavior when the limit is reached? For example,
> 
> * When CPS rate is reached, is it expected that new connections won’t be learned even though the overall CPS per DPU has not maxed out?
> * When CPS or max flows are reached, should a per ENI counter be incremented to track the number of dropped flows due to max limit? Should all or a subset of the dropped flows be sent to the Control Plane/Controller?
> 
> 2. There is a reference to per DPU initial configuration in the routing action comment “list of active modules impacts memory and needs to be configured as part of initial card config”. What other per DPU initial configurations should be expected (e.g. global connection aging time, etc.)?
> 3. For each routing type, please help list a set of valid actions associated with it and any expected action orders in addition to the max number of actions. This info is helpful in writing test cases to validate expected behavior not explicitly listed in the sample json file.
> 4. In term of routing action combinations, Is “maprouting” action the only exception or it is just one example and there are more exception?
> 5. For the “drop” routing-type, what are the other two action types that can be associated with “drop” action-type?
> 6. Please help add more info on what objects can be modified and the expected behaviors. For example, what is the expected behavior for existing connections when  ACL rules change from “allow” to “deny” that impacts those connections?

@lisahnguyen  Moved comments to the correct places in the JSON file.

mmiele:
> 2\. n



> I have a few questions for the examples. I am not able to add them to the corresponding of the document. Hence, I will list them all here
> 
> 1. For the per-ENI’s “qos_settings”, what is the expected behavior when the limit is reached? For example,
> 
> * When CPS rate is reached, is it expected that new connections won’t be learned even though the overall CPS per DPU has not maxed out?
> * When CPS or max flows are reached, should a per ENI counter be incremented to track the number of dropped flows due to max limit? Should all or a subset of the dropped flows be sent to the Control Plane/Controller?
> 
> 2. There is a reference to per DPU initial configuration in the routing action comment “list of active modules impacts memory and needs to be configured as part of initial card config”. What other per DPU initial configurations should be expected (e.g. global connection aging time, etc.)?
> 3. For each routing type, please help list a set of valid actions associated with it and any expected action orders in addition to the max number of actions. This info is helpful in writing test cases to validate expected behavior not explicitly listed in the sample json file.
> 4. In term of routing action combinations, Is “maprouting” action the only exception or it is just one example and there are more exception?
> 5. For the “drop” routing-type, what are the other two action types that can be associated with “drop” action-type?
> 6. Please help add more info on what objects can be modified and the expected behaviors. For example, what is the expected behavior for existing connections when  ACL rules change from “allow” to “deny” that impacts those connections?

@lisahnguyen  Moved comments to the correct places in the JSON file.

prsunny (approved):

mgheorghe:
can we add all elements needed to have the config working please:
- i see interfaces portion missing, where we configure the ip on each interface of the dpu 
- Layer 1 settings are missing also
- loopback interface for vtep is missing also

mmiele:
@mgheorghe This PR used a JSON file as an example. You may want to refer to the current document [SONiC-DASH HLD](https://github.com/Azure/DASH/blob/main/documentation/general/design/dash-sonic-hld.md). Thank you

Issue #88, opened 2022-03-29
IPv6 routing support


KrisNey-MSFT:
## Related information
Scope - P4 DASH pipeline. Requires definition of IPv6 header in overlay pipeline headers, proper parsing, and match on L3 addresses along with IPv4 addresses.

@marian-pritsak to provide further information, give to Hanif.

### Notes
Needs to be coded in P4 behavioral model
No change in bmv2, just P4

mhanif:
Hi @marian-pritsak, Not sure if you saw my email.  As discussed, please send me the information on an equivalent IPv4 model.  Also, please share the development and testing environment details as well.  Thanks!

KrisNey-MSFT:
Adding Marian's email...

425-876-2040 cell
________________________________

marian-pritsak:
Hi Mohammad,

The following list of tables will require IPv6 support:
* sirius_outbound.routing
* sirius_outbound.ca_to_pa
* ACL stages (sirius_acl file)

all those tables match on IPv4 address. We need to support matching on either IPv4 address or IPv6 address depending on the packet's overlay ethertype.

mariobaldi:
We will need IPv6 support also in connection tracking, hence generally throughout sirius_conntrack.p4.

 

My suggestion would be to first finalize connection tracking with the proper connection removal behavior just for IPv4 (building on the current version of the code) and then later add IPv6 support.

PR #97, opened 2022-04-18
IPv6 outbound routing support for overlay processing

mhanif:
This PR is to introduce the outbound IPv6 routing support described in Issue #88.  This addresses couple of P4 tables namely routing_v4/v6 and ca_to_pa_v4/v6.  The ACL support will be added through a separate PR.


sirius-pipeline/README.md
```diff
@@ -13,25 +13,27 @@ make bmv2/sirius_pipeline.bmv2/sirius_pipeline.json
...
+table_add eni_lookup_to_vm inbound.set_eni c:cc:cc:cc:cc:cc => 7
 table_add eni_to_vni set_vni 7 => 9
 table_add routing route_vnet 7 0x01010100/24 => 14
-table_add ca_to_pa set_tunnel_mapping 14 0x01010102 => 0x02020202 88:88:88:88:88:88 1
-table_add ca_to_pa set_tunnel_mapping 14 0x01010103 => 0x02020202 88:88:88:88:88:88 0
+table_add ca_v4_to_pa set_tunnel_mapping 14 0x01010102 => 0x02020202 88:88:88:88:88:88 1
+table_add ca_v4_to_pa set_tunnel_mapping 14 0x01010103 => 0x02020202 88:88:88:88:88:88 0
 table_add appliance set_appliance 0&&&0 => 77:77:77:77:77:77 66:66:66:66:66:66 0x02020201 0
+table_add ca_v6_to_pa set_tunnel_mapping 14 0x01010104 => 0x02020202 88:88:88:88:88:88 1
+table_add ca_v6_to_pa set_tunnel_mapping 14 0x01010105 => 0x02020202 88:88:88:88:88:88 0
```
venkatmahalingam, line 36: Key should be ENI and IPv6 address, please correct it.

sirius-pipeline/bmv2/sirius_outbound.p4
```diff
@@ -53,7 +53,22 @@ control outbound(inout headers_t hdr,
             route_vnet; /* for expressroute - ecmp of overlay */
         }
 
-        counters = routing_counter;
+        counters = routing_v4_counter;
+    }
+
+    direct_counter(CounterType.packets_and_bytes) routing_v6_counter;
+
+    table routing_v6 {
```
marian-pritsak, line 61: @prsunny Do we intend to have separate IPv6 and IPv4 tables or a joint table?

prsunny: We intend to have a joint table.

mhanif: @prsunny Why?  Most of the v4/v6 implementations I have seen for P4 models carry separate v4 and v6 table.  It is just a more cleaner design which allows a more efficient way to maintain/evolve/enhance the code.

KrisNey-MSFT: Table has to decide which key (example: IPv4 or IPv6).  Key construction, parsing the key, pass to table, table executes an action.  Split table to ensure separate buckets to parse (cleaner in one view).  
We have not seen (in SAI APIs) split tables.  We have a v6 and v4 table type in SAI.  

Do we want to have a unified SAI API?  Looking at the IP family will place an entry into the correct table (can place annotation on the table, for example bound to the same API).  Lists are a separate issue...

As an example:  UDP was anchored to same attribute in the metadata.

_After review of SAI, it looks like there is **1 table**, we can follow the model in SAI to remain consistent._

sirius-pipeline/README.md
mhanif: @venkatmahalingam They key is <VNI, IPv6-addr> and it is correctly used <14, 0x01010104>.  By the way, I am making a change such that there is only one table used for both v4 and v6.  Thanks.

sirius-pipeline/bmv2/sirius_outbound.p4
mhanif: @marian-pritsak @prsunny As discussed, I have made the changes to use one table for both v4 and v6 processing.  Please review the changes.  Thanks.

KrisNey-MSFT: @marian-pritsak can likely look at this soon and approve.  Marian would like to merge this first, prior to test infrastructure.

sirius-pipeline/bmv2/sirius_outbound.p4
```diff
@@ -32,7 +32,7 @@ control outbound(inout headers_t hdr,
     table routing {
         key = {
             meta.eni : exact @name("meta.eni:eni");
-            hdr.ipv4.dst_addr : lpm @name("hdr.ipv4.dst_addr:destination");
+            meta.dst_ip_addr : lpm @name("meta.dst_ip_addr:destination");
```
marian-pritsak, line 35: There needs to be a flag to not overlap v4 and v6 prefixes that have identical 4 LSB.

sirius-pipeline/bmv2/sirius_outbound.p4
```diff
@@ -62,7 +62,7 @@ control outbound(inout headers_t hdr,
         key = {
             /* Flow for express route */
             meta.encap_data.dest_vnet_vni : exact @name("meta.encap_data.dest_vnet_vni:dest_vni");
-            hdr.ipv4.dst_addr : exact @name("hdr.ipv4.dst_addr:dip");
+            meta.dst_ip_addr : exact @name("meta.dst_ip_addr:dip");
```
marian-pritsak, line 65: There needs to be a flag to not overlap v4 and v6 prefixes that have identical 4 LSB.

marian-pritsak (changes requested):

sirius-pipeline/bmv2/sirius_pipeline.p4
```diff
@@ -121,6 +121,13 @@ control sirius_ingress(inout headers_t hdr,
             inbound_routing.apply();
         }
 
+        meta.dst_ip_addr = 0;
+        if (hdr.ipv6.isValid()) {
+            meta.dst_ip_addr = hdr.ipv6.dst_addr;
+        } else if (hdr.ipv4.isValid()) {
+            meta.dst_ip_addr = (bit<128>)hdr.ipv4.dst_addr;
```
marian-pritsak, line 128: Upper bits need to be zeroed. I'm not sure if they will be zeroed by default.

sirius-pipeline/bmv2/sirius_outbound.p4
mhanif: @marian-pritsak How do I achieve this in P4?  How do I define this flag?  Do you have some example?  Thanks!

sirius-pipeline/bmv2/sirius_pipeline.p4
mhanif: @marian-pritsak I can force set the upper bits to zero - this should be easy.

sirius-pipeline/bmv2/sirius_outbound.p4
marian-pritsak: Another field bit<1> in metadata, used as match key as well

mhanif: @marian-pritsak Thanks.  I have made the requested changes.  I will update the PR by tonight.

sirius-pipeline/bmv2/sirius_pipeline.p4
mhanif: I have found out from the P4 documentation that, in this case, the upper bits are padded to zeros.  So no changes are required.  Below is an excerpt from the P4 documentation:

<...>
8.9.1. Explicit casts
bit<W> -> bit<X>: truncates the value if W > X, and otherwise (i.e., if W <= X) pads the value with
zero bits.
<...>

sirius-pipeline/bmv2/sirius_outbound.p4
mhanif: @marian-pritsak I have pushed the requested changes to the PR.  Please review.  Thanks!

marian-pritsak (approved):

sirius-pipeline/bmv2/sirius_outbound.p4
KrisNey-MSFT: Marian has a larger merge coming behind this.  This PR takes care of the routing, ACL will come later after Marian's check in  after the new method of creating the environment.

PR #128, opened 2022-06-23
Rename pipeline to DASH

marian-pritsak:


chrispsommers:
OK, as agreed in today's behavioral model WG meeting, let's accept this PR. I will adjust https://github.com/Azure/DASH/pull/127 accordingly. @marian-pritsak also agreed to defer updating READMEs until after PR127 to avoid more churn.

PR #139, opened 2022-06-30
Chris fix CI from PR127

chrispsommers:
Corrected some CI .yml oversights, also reduced strain on CI runners due to ixia-c traffic tests by reducing # packets, packet rate and increasing timeouts.


chrispsommers:
Merging myself since it's just fixing some errors on the initial https://github.com/Azure/DASH/pull/127

Issue #153, opened 2022-07-16
ixia-c tests can fail if protobuf already installed with incompatible version


chrispsommers:
Use of snappi package 0.7.37 as specified in https://github.com/Azure/DASH/blob/main/test/requirements.txt could cause snappi client errors if an incompatible version of protobuf python package is already installed.

Per ixia-c support Slack channel (https://ixia-c.slack.com/archives/C021DU5026R/p1657894469742489?thread_ts=1657847300.692199&cid=C021DU5026R), this can be fixed by upgrading to snappi 0.7.38. I will file a PR to do so.

An example error output is show below:

```
dash@chris-z4:~/chris-dash/DASH/dash-pipeline$ make run-all-tests 
# Ensure P4Runtime server is listening
t=5; \
while [ ${t} -ge 1 ]; do \
	if sudo lsof -i:9559 | grep LISTEN >/dev/null; then \
		break; \
	else \
		sleep 1; \
		t=`expr $t - 1`; \
	fi; \
done; \
docker exec -w /tests/vnet_out simple_switch-dash ./vnet_out
GRPC call SetForwardingPipelineConfig 0.0.0.0:9559 => /etc/dash/dash_pipeline.json, /etc/dash/dash_pipeline_p4rt.txt
GRPC call Write::add_one_entry OK: GRPC call Write::add_one_entry OK: GRPC call Write::add_one_entry OK: Done.
python3 -m pip install -r ../test/requirements.txt
Requirement already satisfied: snappi==0.7.37 in /home/dash/.local/lib/python3.8/site-packages (from -r ../test/requirements.txt (line 1)) (0.7.37)
Requirement already satisfied: pytest==6.0.1 in /home/dash/.local/lib/python3.8/site-packages (from -r ../test/requirements.txt (line 2)) (6.0.1)
Requirement already satisfied: urllib3 in /usr/lib/python3/dist-packages (from snappi==0.7.37->-r ../test/requirements.txt (line 1)) (1.25.8)
Requirement already satisfied: grpcio-tools==1.44.0; python_version > "2.7" in /home/dash/.local/lib/python3.8/site-packages (from snappi==0.7.37->-r ../test/requirements.txt (line 1)) (1.44.0)
Requirement already satisfied: PyYAML in /usr/lib/python3/dist-packages (from snappi==0.7.37->-r ../test/requirements.txt (line 1)) (5.3.1)
[...]

PR #154, opened 2022-07-16
Use snappi 0.7.38

chrispsommers:
Fixes https://github.com/Azure/DASH/issues/153 "ixia-c tests can fail if protobuf already installed with incompatible version"


mgheorghe:
looks good to me

Issue #143, opened 2022-07-14
Dash build bmv2 container permission error


mukeshmv:
The bmv2 container image used in the Dash build requires write access to Volumes mounted on the host machine.
However this docker image comes with hard-coded uid/gid assigned in the Dockerfile when building the image.
The build workflow involves using a pre-built bmv2 image pulled from the registry.
When running a container from this docker image, the build commands run from within the container using this uid try to write to the mounted volumes in the host machine resulting in permission issues.
 
```
# make sai
Generate SAI library headers and implementation...
docker run -v /home/mukesh/dash/dash-pipeline/bmv2:/bmv2 -v /home/mukesh/dash/dash-pipeline/SAI:/SAI -v /home/mukesh/dash/dash-pipeline/tests:/tests --network=host --rm -it \
        --name build_sai-mukesh \
        -w /SAI chrissommers/dash-bmv2:pr127-220623 \
    ./generate_dash_api.sh
Directory ./lib will be deleted...
Traceback (most recent call last):
  File "./sai_api_gen.py", line 357, in <module>
    shutil.rmtree('./lib')
  File "/usr/lib/python3.8/shutil.py", line 718, in rmtree
    _rmtree_safe_fd(fd, path, onerror)
  File "/usr/lib/python3.8/shutil.py", line 675, in _rmtree_safe_fd
    onerror(os.unlink, fullname, sys.exc_info())
  File "/usr/lib/python3.8/shutil.py", line 673, in _rmtree_safe_fd
    os.unlink(entry.name, dir_fd=topfd)
PermissionError: [Errno 13] Permission denied: 'utils.cpp'
Makefile:85: recipe for target 'sai' failed
```

mukeshmv:
Workaround -  
```
cd dash-pipeline
chmod -R o+rw SAI/ bmv2/ tests/
```

chrispsommers:
Thanks for filing this! I'll work on a permanent fix.

PR #155, opened 2022-07-17
Temporary workaround for container permissions issues 

chrispsommers:
Fixes https://github.com/Azure/DASH/issues/143.
Executes `chmod `as required. Permanent fix will require some Docker mods.
Note, instead of `make clean all` please use `make clean && make all` to ensure `chmod` gets called prior to both make targets.

Reviewers, please clone the dev branch and try fresh `make clean && make all network run-switch` to validate this works for other users. I've tested as another alias user and it worked, but I prefer independent testing. Thanks!

### DIFF
--- a/dash-pipeline/Makefile
+++ b/dash-pipeline/Makefile
@@ -1,6 +1,6 @@
 SHELL = /bin/bash
 
-all:p4 sai test
+all:fix-perms p4 sai test
 
 PWD := $(shell pwd)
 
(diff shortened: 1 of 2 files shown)


mukeshmv:
Works for me.

mukeshmv (approved):

chrispsommers:
Thanks @budgrise and @mukeshmv !

PR #160, opened 2022-07-21
Add support for IPV4 optional headers

saad-mzhr:
**Summary of PR**
Right now, if you send an ipv4 packet with an optional header to the pipeline, the parser will fail because it will try to extract UDP/TCP immediately after plain ipv4 header. This PR adds support for IPv4 options in the parser/deparser.

**How it was tested?** @chrispsommers 
1. Since this changeset is only related to parser/deparser, I temporarily emptied out the dash-pipeline such that is simply sends every packet coming from port 0 out to port 1
2. Bring up the setup and send different packets with/without ipv4 options to veth0
3. Examine the logs of simple_switch to verify the functionality of parser
4. Examine the contents of packets coming out of veth2 to verify the functionality of deparser

### DIFF
--- a/dash-pipeline/bmv2/dash_headers.p4
+++ b/dash-pipeline/bmv2/dash_headers.p4
@@ -31,6 +31,10 @@ header ipv4_t {
 
 const bit<16> IPV4_HDR_SIZE=160/8;
 
+header ipv4options_t {
+    varbit<320> options;
+}
+
 header udp_t {
     bit<16>  src_port;
     bit<16>  dst_port;
@@ -81,6 +85,7 @@ const bit<16> IPV6_HDR_SIZE=320/8;
 struct headers_t {
     ethernet_t ethernet;
     ipv4_t     ipv4;
+    ipv4options_t ipv4options;
     ipv6_t     ipv6;
     udp_t      udp;
     tcp_t      tcp;
--- a/dash-pipeline/bmv2/dash_parser.p4
+++ b/dash-pipeline/bmv2/dash_parser.p4
@@ -5,7 +5,8 @@
 
 error {
     IPv4IncorrectVersion,
-    IPv4OptionsNotSupported
+    IPv4OptionsNotSupported,
+    InvalidIPv4Header
 }
 
 #define UDP_PORT_VXLAN 4789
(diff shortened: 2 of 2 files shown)


chrispsommers:
LGTM

dash-pipeline/bmv2/dash_headers.p4
```diff
@@ -81,6 +85,7 @@ const bit<16> IPV6_HDR_SIZE=320/8;
 struct headers_t {
     ethernet_t ethernet;
     ipv4_t     ipv4;
+    ipv4options_t ipv4options;
```
venkatmahalingam, line 88: Don't we need to move ipv4options_t part of ipv4_t itself?

chrispsommers:
Maybe write a test-case? You should be able to send some packets in and out w/ current pipeline. Even with incomplete pipeline code right now, packets will all come back out port 0.

dash-pipeline/bmv2/dash_headers.p4
chrispsommers: @saad-mzhr Could you cite the example you based this on from the P4-16 lang spec? You can post the URL to the appropriate line or section in the HTML version of the spec. Thanks.

saad-mzhr: Maybe. I actually adapted this code from p4 official docs on how to handle variable sized headers:
https://p4.org/p4-spec/docs/P4-16-v1.0.0-spec.html#sec-packet-extract-two

saad-mzhr:
> Maybe write a test-case? You should be able to send some packets in and out w/ current pipeline. Even with incomplete pipeline code right now, packets will all come back out port 0.

Sure

saad-mzhr:
> LGTM

Whats "LGTM"? :)

chrispsommers:
LGTM = common shorthand for "Looks Good to me." It's a common GitHub way to say "I approve." I had to learn the lingo too, I remember the first time I saw this not too long ago. :)

dash-pipeline/bmv2/dash_headers.p4
chrispsommers: Personally, I like this approach of treating the extensions header separately, since it's an optional part of the header stack.

venkatmahalingam: IPv4 header is basically including options, so, I suggest to rename ipv4_t with IPv4_no_options_t

mhanif:
LGTM

Issue #158, opened 2022-07-21
DASH libsai fails on remove() operations, wrong log message


chrispsommers:
I modified vnet_out.cpp to delete items added in the test and revealed some problems. Here's the added SAI code:
```

    // Delete everything in reverse order
    status = sai_remove_outbound_eni_to_vni_entry(&e2v);
    if (status != SAI_STATUS_SUCCESS)
    {
        std::cout << "Failed to remove ENI To VNI" << std::endl;
        return 1;
    }

    status = sai_remove_eni_ether_address_map_entry(&eam);
    if (status != SAI_STATUS_SUCCESS)
    {
        std::cout << "Failed to remove ENI Lookup From VM" << std::endl;
        return 1;
    }

    status = sai_remove_direction_lookup_entry(&dle);
    if (status != SAI_STATUS_SUCCESS)
    {
        std::cout << "Failed to remove Direction Lookup Entry" << std::endl;
        return 1;
    }
```
Here's the output from bmv2 switch. We see three creates and one delete (which succeeded), there should be three deletes but the test aborted.
```
[21:37:46.092] [bmv2] [D] [thread 25] Entry 0 added to table 'dash_ingress.direction_lookup|dash'
[21:37:46.092] [bmv2] [D] [thread 25] Dumping entry 0
Match key:
* hdr.vxlan.vni:VNI   : EXACT     3c0000
Action entry: dash_ingress.set_outbound_direction - 

[21:37:46.092] [bmv2] [D] [thread 26] Entry 0 added to table 'dash_ingress.eni_ether_address_map|dash'
[21:37:46.092] [bmv2] [D] [thread 26] Dumping entry 0
Match key:
* meta.eni_addr:address: EXACT     aacccccccccc
Action entry: dash_ingress.set_eni - 7,

[21:37:46.093] [bmv2] [D] [thread 25] Entry 0 added to table 'dash_ingress.outbound.eni_to_vni|dash_vnet'
[21:37:46.093] [bmv2] [D] [thread 25] Dumping entry 0
Match key:
* meta.eni_id:eni_id  : EXACT     0007
Action entry: dash_ingress.outbound.set_vni - 90000,

[21:37:46.093] [bmv2] [D] [thread 26] Removed entry 0 from table 'dash_ingress.outbound.eni_to_vni
```
Here's the console output when running vnet_out:
```
$ make run-test 
...
[...]

PR #161, opened 2022-07-22
Fix issue 158

chrispsommers:
Fix https://github.com/Azure/DASH/issues/158 - libsai delete operation failure and log msg. The test for error code was backwards and log message used same write updateType string for all operations. I fixed the compare and used a dynamic enum print method.


marian-pritsak:
Can we add the exact description of the fix instead of issue #? It's impossible to track in git.

chrispsommers:
> Can we add the exact description of the fix instead of issue #? It's impossible to track in git.

@marian-pritsak  Sorry, the commit message didn't display. Please see revised description.

chrispsommers:
Fixed in https://github.com/Azure/DASH/pull/181, no need for this PR any more.

PR #171, opened 2022-08-03
DASH SAI Packet Testing Framework Design HLD

reshmaintel:
DASH SAI Packet Testing Framework Design HLD


doc_SAI-Proposal-SAI-PTF.md
```diff
@@ -0,0 +1,723 @@
+# SAI PTF
```
chrispsommers, line 1: I think this file should not be uploaded to DASH since it's going to live under SAI. Could you instead make another entry in the table under https://github.com/Azure/DASH/tree/main/test/docs#table-of-contents---dash-test-documentation like you did for the SAI-PTF User guide? Once your proposal is upstreamed we can change the URL to point there. For now it can point to your dev branch. Thanks!

chrispsommers (changes requested):
Slides are OK, but SAI PTF proposal file should be changed to an external link.

doc_SAI-Proposal-SAI-PTF.md
KrisNey-MSFT: @reshmaintel to review

vmytnykx: Added reference to design instead of the doc itself.

test/docs/README.md
```diff
@@ -5,6 +5,7 @@
 | [High-Level Description (HLD) Test Specification](dash-test-HLD.md) | High-level design for the testing of devices which conform to the SONiC-DASH requirements.|  
 | [Dash Test Maturity Stages](dash-test-maturity-stages.md) | Describes a progressive approach to DASH testing.|  
 | [DASH SAI-Thrift Test Workflow](dash-test-workflow-saithrift.md) | DASH test workflow with SAI-thrift. |
+| [SAI PTF Design](https://github.com/reshmaintel/SAI/blob/dash-ptf/doc/SAI-Proposal-SAI-PTF.md) | SAI Thrift auto-generated Python based testing framework doc. |
```
chrispsommers, line 8: Please use upstream URL: https://github.com/reshmaintel/SAI/blob/dash-ptf/doc/SAI-Proposal-SAI-PTF.md

chrispsommers (changes requested):
Please use upstream URL, thanks

test/docs/README.md
vmytnykx: @chrispsommers the modified design doc is not upstreamed yet, so based on https://github.com/Azure/DASH/pull/171#discussion_r937058503 comment, I've added dev branch for now. Later it can be updated to up streamed one.

PR #172, opened 2022-08-03
Remove 'nh_' from direct route's overlay_nh_ip

marian-pritsak:


marian-pritsak:
Removed per @prsunny request

prsunny (approved):

PR #176, opened 2022-08-04
Move underlay SIP resolution to VIP table

marian-pritsak:


marian-pritsak:
Does not change SAI API

Issue #147, opened 2022-07-14
Improve drop handling


budgrise:
- [x] Need to have a consistent plan for packet drops. 
When we decide to drop a packet, do we just mark it for drop or stop processing immediately with a "return"?
Also need to consider when counters are incremented.

- [ ] The vip table's deny action is logically @defaultonly. 
But making this change would break the SAI generation due to the generation tool.
If the SAI generation tool is enhanced, then we could make this change.

- [x] The lookup for table eni_ether_address_map should drop the packet on lookup miss.
In general, need more negative checking in the code.

PR #187, opened 2022-08-09
Default drop support for P4 table miss without generating SAI drop action

### LINKED ISSUE #177: P4 Issue: egress port should = ingress port

mukeshmv:
Fixes part of #147
Handle all table miss drops consistently at the end of the pipeline

Fixes #177 

SAI API change - Removes the following action
`
SAI_INBOUND_ROUTING_ENTRY_ACTION_DENY 
`

Could not added defaultonly tag to the drop action in the vip and pa_validation tables as this removes all SAI attributes for these APIs. SAI does not support apis with no attributes. Ran into the following metadata check errors when this was attempted
```
ERROR: enum sai_vip_entry_attr_t is empty, after removing suffixed entries _START/_END/_RANGE_BASE
ERROR: enum sai_pa_validation_entry_attr_t is empty, after removing suffixed entries _START/_END/_RANGE_BASE
```
or this when I removed the sai_xxx_attr_t enum completely.
```
ERROR: SAI_OBJECT_TYPE_VIP_ENTRY is not defined in OBJTOAPIMAP, missing sai_XXX_api_t declaration?
ERROR: SAI_OBJECT_TYPE_PA_VALIDATION_ENTRY is not defined in OBJTOAPIMAP, missing sai_XXX_api_t declaration?
```

### DIFF
--- a/dash-pipeline/bmv2/dash_pipeline.p4
+++ b/dash-pipeline/bmv2/dash_pipeline.p4
@@ -140,7 +140,9 @@ control dash_ingress(inout headers_t hdr,
 
         actions = {
             set_eni_attrs;
+            @defaultonly deny;
         }
+        const default_action = deny;
     }
 
     direct_counter(CounterType.packets_and_bytes) eni_counter;
@@ -158,7 +160,6 @@ control dash_ingress(inout headers_t hdr,
     }
 
     action permit() {
-        meta.dropped = false;
     }
 
     action vxlan_decap_pa_validate(bit<16> src_vnet_id) {
@@ -174,7 +175,7 @@ control dash_ingress(inout headers_t hdr,
 
         actions = {
             permit;
-            @defaultonly deny;
+            deny;
         }
 
         const default_action = deny;
@@ -208,7 +209,9 @@ control dash_ingress(inout headers_t hdr,
 
         actions = {
             set_eni;
+            @defaultonly deny;
         }
+        const default_action = deny;
     }
 
     action set_acl_group_attrs(bit<32> ip_addr_family) {
@@ -242,12 +245,6 @@ control dash_ingress(inout headers_t hdr,
             /* Use the same VIP that was in packet's destination if it's
                present in the VIP table */
             meta.encap_data.underlay_sip = hdr.ipv4.dst_addr;
-        }
-        // TODO [cs] shouldn't this also be called at end of ingress?
-        // Shouldn't it call mark_to_drop(standard_metadata);
-
-        if (meta.dropped) {
-            return;
         }
 
         /* If Outer VNI matches with a reserved VNI, then the direction is Outbound - */
@@ -288,7 +285,6 @@ control dash_ingress(inout headers_t hdr,
         eni.apply();
         if (meta.eni_data.admin_state == 0) {
             deny();
-            return;
         }
         acl_group.apply();
 
--- a/dash-pipeline/tests/saithrift/ptf/vnet/test_saithrift_vnet.py
+++ b/dash-pipeline/tests/saithrift/ptf/vnet/test_saithrift_vnet.py
@@ -126,63 +126,62 @@ class TestSaiThrift_outbound_udp_pkt(ThriftInterfaceDataPlane):
             outer_smac = "00:00:05:06:06:06"
             inner_smac = "00:00:04:06:06:06"
 
-            # TODO - Enable drop tests once PR# 187 is merged
-            ## check VIP drop
-            #wrong_vip = "172.16.100.100"
-            #inner_pkt = simple_udp_packet(eth_dst="02:02:02:02:02:02",
-            #                              eth_src=self.eni_mac,
-            #                              ip_dst=self.dst_ca_ip,
-            #                              ip_src=src_vm_ip)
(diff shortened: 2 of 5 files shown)


dash-pipeline/tests/saithrift/ptf/vnet/test_saithrift_vnet.py
```diff
@@ -94,9 +94,10 @@ def runTest(self):
         # TODO expected packet might be different
         self.udp_pkt_exp = self.udp_pkt
         print("\nSending packet...", self.udp_pkt.__repr__())
-        send_packet(self, 0, self.udp_pkt)
-        print("\nVerifying packet...", self.udp_pkt_exp.__repr__())
-        verify_packet(self, self.udp_pkt_exp, 0)
+        #send_packet(self, 0, self.udp_pkt)
```
chrispsommers, line 97: I'm not in favor of removing any test of pipeline continuity. I'd much prefer we configure the pipeline even with basic wildcard match rules, keep the echo test, then have a test which proves no packets make it without config. (I understand this echo test was relying on incorrect behavior but I don't want to lose some verification.)

dash-pipeline/tests/saithrift/pytest/echo/test_echo_port.py
```diff
@@ -3,6 +3,7 @@
 
 
 @pytest.mark.bmv2
+@pytest.mark.skip(reason="re-enable once pipeline fwd is fixed")
```
chrispsommers, line 6: Same comments as above.

dash-pipeline/tests/saithrift/ptf/vnet/test_saithrift_vnet.py
mukeshmv: I am going to be creating a follow up PR immediately to convert this to an actual dash vxlan outbound packet test. Already have all the changes ready. Just waiting for the ptf vxlan fix.

dash-pipeline/tests/saithrift/pytest/echo/test_echo_port.py
mukeshmv: i am not sure about this one, need to check with you and Marian on the functionality being tested here and how we can convert this. 
Since with this PR, any table miss leads to actual packet drop, so these tests fail now.

dash-pipeline/tests/saithrift/ptf/vnet/test_saithrift_vnet.py
mukeshmv: Changed this to a negative test to verify invalid packet drop

dash-pipeline/bmv2/dash_pipeline.p4
```diff
@@ -295,9 +291,6 @@ control dash_ingress(inout headers_t hdr,
 
         if (meta.dropped) {
             drop_action();
-        } else {
-            /* Send packet to port 1 by default if we reached the end of pipeline */
-            standard_metadata.egress_spec = 1;
```
chrispsommers, line 300: Would be nice to loop it back (`standard_metadata.egress_spec = stanard_metadata.ingress_port`). Deleting this assignment will send all packets to egress_spec=0 by default in bmv2.

dash-pipeline/tests/saithrift/ptf/vnet/test_saithrift_vnet.py
```diff
@@ -90,13 +90,12 @@ def setUp(self):
...
         self.udp_pkt = simple_udp_packet()
-        # TODO expected packet might be different
-        self.udp_pkt_exp = self.udp_pkt
-        print("\nSending packet...", self.udp_pkt.__repr__())
+        print("\nSending packet to invalid vip...", self.udp_pkt.__repr__())
         send_packet(self, 0, self.udp_pkt)
-        print("\nVerifying packet...", self.udp_pkt_exp.__repr__())
-        verify_packet(self, self.udp_pkt_exp, 0)
+        print("\nVerifying packet drop...")
+        verify_no_other_packets(self, 0)
```
chrispsommers, line 98: Note in some environments this might fail because of IPv67 control-plane packets etc. bombarding the veths. So if this starts failing in people's dev environments that might be why.

chrispsommers (approved):

dash-pipeline/bmv2/dash_pipeline.p4
mukeshmv: @chrispsommers yes it is already looped back by default at the beginning of the pipeline at line dash_pipeline.c : 232. I just removed this line so that it takes effect.

chrispsommers: @mukeshmv perfect, I missed that. Thanks!

chrispsommers:
LGTM!

dash-pipeline/tests/saithrift/ptf/vnet/test_saithrift_vnet.py
chrispsommers: OK.

dash-pipeline/tests/saithrift/pytest/echo/test_echo_port.py
chrispsommers: ok

marian-pritsak (approved):

PR #193, opened 2022-08-11
Dash outbound pipeline packet test

mukeshmv:
- Fix issues in #185
- Upgrade to p4c container image with fix for https://github.com/p4lang/p4c/issues/3488

PR #194, opened 2022-08-11
Fix inbound VNET routing classification

marian-pritsak:
Classify inbound packet to VNET instead of using VNI
as VNIs alone may overlap

### DIFF
--- a/dash-pipeline/bmv2/dash_pipeline.p4
+++ b/dash-pipeline/bmv2/dash_pipeline.p4
@@ -161,14 +161,15 @@ control dash_ingress(inout headers_t hdr,
         meta.dropped = false;
     }
 
-    action vxlan_decap_pa_validate() {}
+    action vxlan_decap_pa_validate(bit<16> src_vnet_id) {
+        meta.vnet_id = src_vnet_id;
+    }
 
     @name("pa_validation|dash_vnet")
     table pa_validation {
         key = {
-            meta.eni_id: exact @name("meta.eni_id:eni_id");
+            meta.vnet_id: exact @name("meta.vnet_id:vnet_id");
             hdr.ipv4.src_addr : exact @name("hdr.ipv4.src_addr:sip");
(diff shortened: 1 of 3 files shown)


chrispsommers (changes requested):
Please write a PTF or Pytest test-case to validate the behavior, thanks.

dash-pipeline/bmv2/dash_pipeline.p4
```diff
@@ -182,7 +183,9 @@ control dash_ingress(inout headers_t hdr,
     @name("inbound_routing|dash_vnet")
     table inbound_routing {
         key = {
+            meta.eni_id: exact @name("meta.eni_id:eni_id");
             hdr.vxlan.vni : exact @name("hdr.vxlan.vni:VNI");
+            hdr.ipv4.src_addr : ternary @name("hdr.ipv4.src_addr:sip");
```
mukeshmv, line 188: So there is a src_addr in both the inbound_routing and pa_validation. So even if the inbound routing rule is for a specific src_addr we still expect a pa_validation entry for the same ?

prsunny: @vmytnykx , inbound routing is not LPM - Please refer here - https://github.com/Azure/DASH/blob/main/documentation/general/design/dash-sonic-hld.md#22-inbound-packet-processing-pipeline. @KrisNey-MSFT , we may want to correct the vnet-to-vnet-service.md section.

marian-pritsak:
> Please write a PTF or Pytest test-case to validate the behavior, thanks. Needs #187 (drop behavior) and #193 (PTF support for VXLANS).

I don't think this commit should include a test. It's not even a new table. My primary goal is to get the VNET definition to a steady state, where we don't expect any more changes so that tests that follow won't need to be updated every time.

chrispsommers (approved):

PR #202, opened 2022-08-24
Fix docker & Makefile perms

chrispsommers:
**NOTE**: prior `make all`'s will leave content in your workspace which can't be `clean`ed using the modified makefiles, so please `make clean` before upgrading to this. If necessary, use `sudo rm -rf <dangling content>` to clean up manually.

- Remove all `sudo` commands from Makefiles except those which use Linux netdevs during SW packet tests.
- More selective use of `chmod` commands (w/o `sudo`) during make, to grant access to running containers to modify the user workspace via volume mounts
- Refactor all Docker images to run as user `dashuser:dashuser` instead of original Docerfile author's identify (moi).
- Change CI script to run as normal CI user (`runner`) except for SW packet tests, to replicate the developer experience
- Test for `veth`'s before deleting them to avoid superfluous deletes which require `sudo`


chrispsommers:
Reviewers, please clone the dev branch and see if you can build and run tests. Thanks!

mukeshmv (approved):
lgtm. 
Worked for me. Faced some permission issues with the first make clean on existing workspace. But that was most probably my bad since I should have cleaned first before switching to this branch. But subsequent repeated make clean and make all went through fine. Followed by successful PTF test run.

chrispsommers:
> lgtm. Worked for me. Faced some permission issues with the first make clean on existing workspace. But that was most probably my bad since I should have cleaned first before switching to this branch. But subsequent repeated make clean and make all went through fine. Followed by successful PTF test run.

@mukeshmv Thanks for trying! I had the same issue myself just today, I'll add comments to the PR.

vmytnykx:
@chrispsommers  it works for me. Just one note probably not related to this changes:

Had to modify `test/third-party/traffic_gen/[un]deploy_ixiac.sh` scripts to use `docker compose` instead of `docker-compose` because of `docker-compose: command not found` error. I think we might need to move those scripts to new docker compose tool instead of deprecated  `docker-compose`.

vmytnykx (approved):

chrispsommers:
> @chrispsommers it works for me. Just one note probably not related to this changes:
> 
> Had to modify `test/third-party/traffic_gen/[un]deploy_ixiac.sh` scripts to use `docker compose` instead of `docker-compose` because of `docker-compose: command not found` error. I think we might need to move those scripts to new docker compose tool instead of deprecated `docker-compose`.

@vmytnykx Thanks for the review! re: `docker- compose` vs. `docker compose` - I wasn't aware of the new command. I'll look into reconciling this. Meanwhile, did you follow the instructions [here](https://github.com/Azure/DASH/blob/main/dash-pipeline/README.md#install-docker-compose)? I'm not disagreeing that we should update to use `docker compose` but I want to make sure there's no functional problems with the existing workflow. Thanks.

vmytnykx:
> > @chrispsommers it works for me. Just one note probably not related to this changes:
> > Had to modify `test/third-party/traffic_gen/[un]deploy_ixiac.sh` scripts to use `docker compose` instead of `docker-compose` because of `docker-compose: command not found` error. I think we might need to move those scripts to new docker compose tool instead of deprecated `docker-compose`.
> 
> @vmytnykx Thanks for the review! re: `docker- compose` vs. `docker compose` - I wasn't aware of the new command. I'll look into reconciling this. Meanwhile, did you follow the instructions [here](https://github.com/Azure/DASH/blob/main/dash-pipeline/README.md#install-docker-compose)? I'm not disagreeing that we should update to use `docker compose` but I want to make sure there's no functional problems with the existing workflow. Thanks.

Yes, it works also. But I think in the future we will need to avoid deprecated `docker- compose`. For now changes are good.

KrisNey-MSFT:
Merge 201, then this one

marian-pritsak (approved):

KrisNey-MSFT:
Per Community conversations and PR reviews, merging.

PR #206, opened 2022-08-31
AMD-Pensando HA Proposal document

SanjayTh:


documentation/high-avail/design/AMD-Pensando_HA_Proposal.md
```diff
@@ -0,0 +1,723 @@
...
+
+The sync packets between the DPUs are logically structured as below.
+
+![](images/dp_repl_packet.005.png)
+
+### Interaction between Bulk Sync and Datapath Sync
+
+Due to the scale requirements for DASH the flow table size that needs to be handled during bulk sync can be very large and hence the bulk sync process can take a long time to complete. It is not possible to halt all traffic that would create new flows during this time. Hence the sync mechanism has to handle creation of new flows during bulk sync. It is also possible that there might be changes in the policy that might affect existing flows. The perfect sync mechanism calls for marking different  “color”s to flows that are created after the start of bulk sync. The flow table is walked and all flows not the current color are synchronized to the peer. Any flows that are created during the bulk sync phase are inline synchronized via the datapath synchronization path. Other challenges include
+
```
lguohan, line 95: > It is also possible that there might be changes in the policy that might affect existing flows. The perfect sync mechanism calls for marking different  “color”s to flows that are created after the start of bulk sync.

can different what are the state (color) can we mark for the flow in hardware. can we get specific state listed.

documentation/high-avail/design/AMD-Pensando_HA_Proposal.md
```diff
@@ -0,0 +1,723 @@
...
+### Datapath Heartbeat
+
+Each DPU sends heartbeat messages at a configured interval to its peer. When a peer loses a set number of heartbeats it declares the peer unreachable and a switchover is initiated. The interval between heartbeats and the number of missed heartbeats are configurable. The Heartbeats can be aggressive hence it is left to the DPU to perform the heartbeat.
+
+## State Synchronization
+
+State synchronization between the 2 DPUs uses the CNIP IP. All state synchronization happens at the granularity of the DP-VIP and happens from the primary of the DP-VIP towards the secondary. State synchronization happens in 2 stages
```
lguohan, line 57: is the state synchronization happening in stages or in parallel? would be great to clarify this.

documentation/high-avail/design/AMD-Pensando_HA_Proposal.md
SanjayTh: At any given time during bulk sync, there are flows that are in the bulk sync snapshot and flows that are were created after the bulk sync started. These classes of flows can be differentiated by the "color". Apart from this within the flows that are in the bulk sync snapshot, there are flows that have already been processed and flows that are yet to be processed. It is okay to rely DP sync for the former but not for the later. The implementation may need to track this in the state and handle. Will add this to the description.

documentation/high-avail/design/AMD-Pensando_HA_Proposal.md
SanjayTh: Yes both are happening in parallel. Will state it here explicitly.

KrisNey-MSFT: opened PR209 to address this [PR209](https://github.com/Azure/DASH/pull/209)

documentation/high-avail/design/AMD-Pensando_HA_Proposal.md
```diff
@@ -0,0 +1,723 @@
...
+
+typedef struct _sai_dash_ha_api_t {
+    sai_register_cp_channel_fn          register_cp_channel;
+    sai_get_capabilities_fn             get_capabilities;
+    sai_process_peer_capabilities_fn    process_peer_capabilities;
+    sai_process_dpu_control_message_fn  process_dpu_control_message;
+    sai_process_flow_sync_message_fn    process_flow_sync_message;
+    sai_oper_role_status_fn             oper_role_status;
+    sai_cp_control_message_fn           cp_control_message;
+} sai_dash_ha_api_t;
```
lguohan, line 480: can we have sample code how to call these APIs?

PR #207, opened 2022-08-31
SAI apigen support for tables with no action parameters and a single action

### LINKED ISSUE #147: Improve drop handling

mukeshmv:
Fixes #147 task 2 allowing to add @defaultonly tag to vip and pa_validation table deny actions

SAI api header change - removes the following actions
```
    SAI_VIP_ENTRY_ACTION_DENY
    SAI_PA_VALIDATION_ENTRY_ACTION_DENY
```

### DIFF
--- a/dash-pipeline/SAI/templates/saiapi.h.j2
+++ b/dash-pipeline/SAI/templates/saiapi.h.j2
@@ -34,7 +34,7 @@
  */
 
 {% for table in sai_api.tables %}
-{% if table.actions | length > 1 %}
+{% if table.actions | length > 1 or ((table.actions | length == 1) and (((table.is_object == 'false') or (table['keys'] | length <= 1)) and ((table['actionParams'] | length == 0))))  %}
 /**
  * @brief Attribute data for #SAI_{{ table.name | upper }}_ATTR_ACTION
  */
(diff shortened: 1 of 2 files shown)


marian-pritsak (approved):

dash-pipeline/SAI/templates/saiapi.h.j2
```diff
@@ -34,7 +34,7 @@
  */
 
 {% for table in sai_api.tables %}
-{% if table.actions | length > 1 %}
+{% if table.actions | length > 1 or ((table.actions | length == 1) and (((table.is_object == 'false') or (table['keys'] | length <= 1)) and ((table['actionParams'] | length == 0))))  %}
```
mukeshmv, line 37: Original condition tries to generate action-type even for sai object if it has only 1 action with 0 params. Also it skips 0 action case. If these are no longer needed / not possible, then this simpler should be fine.

Issue #205, opened 2022-08-30
make run-all-tests fails (sending with wrong VIP and verifying drop FAILS)


mhanif:
Just cloned a fresh copy of DASH at the tip of master branch.  Built and tested (unmodified) to get the baseline and saw that it is failing:

$ make run-saithrift-server
docker run --rm -it --net=host --name dash-saithrift-server-mhanif -v /home/mhanif/sw/dash-ipv6/DASH/dash-pipeline/bmv2/dash_pipeline.bmv2/dash_pipeline.json:/etc/dash/dash_pipeline.json -v /home/mhanif/sw/dash-ipv6/DASH/dash-pipeline/bmv2/dash_pipeline.bmv2/dash_pipeline_p4rt.txt:/etc/dash/dash_pipeline_p4rt.txt -v /home/mhanif/sw/dash-ipv6/DASH/dash-pipeline/SAI:/SAI -v /home/mhanif/sw/dash-ipv6/DASH/dash-pipeline/SAI/SAI/meta:/meta -w /SAI/rpc/usr/sbin -e LD_LIBRARY_PATH=/SAI/lib:/usr/local/lib chrissommers/dash-saithrift-bldr:220719 \
./saiserver
GRPC call SetForwardingPipelineConfig 0.0.0.0:9559 => /etc/dash/dash_pipeline.json, /etc/dash/dash_pipeline_p4rt.txt
Starting SAI RPC server on port 9092
GRPC call Write::INSERT OKtable_id: 45621839 match { field_id: 1 exact { value: "\254\020\001d" } } action { action { action_id: 26041632 } }
GRPC call Write::INSERT OKtable_id: 41103078 match { field_id: 1 exact { value: "\000\000d" } } action { action { action_id: 21912829 } }
GRPC call Write::INSERT OKtable_id: 45323240 match { field_id: 1 exact { value: "\000\001" } } action { action { action_id: 25655048 params { param_id: 1 value: "\000\000\000\000" } } }
GRPC call Write::INSERT OKtable_id: 45323240 match { field_id: 1 exact { value: "\000\002" } } action { action { action_id: 25655048 params { param_id: 1 value: "\000\000\000\000" } } }
GRPC call Write::INSERT OKtable_id: 34579306 match { field_id: 1 exact { value: "\000\003" } } action { action { action_id: 17187022 params { param_id: 1 value: "\000\000d" } } }
[...]

PR #211, opened 2022-09-02
Fix ipv6 packet noise

chrispsommers:
Fixes https://github.com/Azure/DASH/issues/205 (make run-all-tests fails).
- add `disable-ipv6` make target dependencies prior to running PTF and Pytests as stop-gap to undo the automatic enabling of IPv6 when ixia-c starts.
- some minor refactoring of related make targets for tidiness
- also changed docker registry from which we pull ixia-c, it's moved from Dockerhub to Google Cloud. First build in any environment will re-pull the images one time.

Root cause was described in https://github.com/Azure/DASH/issues/205#issuecomment-1234534247. I am working with ixia-c dev team to learn why launching ixia-c traffic-generator re-enables IPv6 automatically.


KrisNey-MSFT:
Marian and Community ok with this fix (fixes issue of PTF test potentially failing due to IPv6 Linux (discovery) control packet becoming re-enabled and insertion into PTF test when it was expecting (for example) a different response such as a modified VXLan packet.  
Idea here is to not filter unexpected packets; rather disable IPv6 instead. 

If others know more about name-spacing, please pitch in to help isolate. 

We can discuss the long-term fix offline (as IPv6 control packets are valid packets, however not expected at this point in the test).

mukeshmv:
@chrispsommers I tried it out. I think it fixes the issue from the PTF test setup expectation point of view. 

One observation though - when I run the following sequence
```
make kill-all
make run-switch
make run-saithrift-server
make run-all-tests
```
I see that at the end the ipv6 is enabled
```
sysctl net.ipv6.conf.veth0.disable_ipv6
net.ipv6.conf.veth0.disable_ipv6 = 0
```
But if I then again issue a 
`make run-all 
`
Now I see that at the end ipv6 is disabled
```
sysctl net.ipv6.conf.veth0.disable_ipv6
net.ipv6.conf.veth0.disable_ipv6 = 1
```

I am assuming that this is because the deploy-ixiac is not run the second time so it remains disabled after the ptf tests. 
But I thought your intention was to disable ipv6 immediately even the first time after deploy-ixiac runs. But its not.

chrispsommers:
@mukeshmv Thanks for reporting, I will investigate, It might be due to the Makefile dependency analysis which sometimes doesn't execute targets if it determines they're already "resolved." So, for example, if there is a chain of targets which each have a dependency and it executes the dependency on the first target, it might not execute it for a later target. I'll try to reproduce.

marian-pritsak (approved):

chrispsommers:
@mukeshmv The Makefile was working as intended; I may have verbally miscommunicated the nature of "fix"  in the WG meeting, but the PR description in fact says:
>add disable-ipv6 make target dependencies prior to running PTF and Pytests as stop-gap to undo the automatic enabling of IPv6 when ixia-c starts.

That being said, I just made a minor change to proactively disable IPv6 immediately after `deploy-ixiac` as another safeguard and reduce the window of IPv6 noise. Also in the next week or two, ixia-c will have a new release which allows IPV6 to be disabled at startup.

PR #226, opened 2022-09-16
Upgrade to ixia-c 0.0.1-3383

chrispsommers:
Fixes flow stats for hairpin mode and adds option to disable IPv6 at launch.

Remove ./disable_veth_ipv6.sh from makefile after deploy-ixiac since it's not needed. (Keep it as prerequisite to running test-cases since we still need to prevent linux ICMP6 packets during tests.)


test/third-party/traffic_gen/deployment/.env
```diff
@@ -1,9 +1,10 @@
 DOCKER_REGISTRY=ghcr.io/open-traffic-generator
-CONTROLLER_VERSION=0.0.1-2934
+CONTROLLER_VERSION=0.0.1-3383
 TRAFFIC_ENGINE_VERSION=1.4.1.26
```
bortok, line 3: You need to bump TE version to match Controller. See [release notes](https://github.com/open-traffic-generator/ixia-c/releases/tag/v0.0.1-3383) for compatible versions.

bortok (changes requested):
Chris, you also need to update snappi version in the requirements.txt to match controller ver

test/third-party/traffic_gen/deployment/.env
chrispsommers: Thanks, I just pushed this change. Glad you caught it!

bortok (approved):

PR #234, opened 2022-09-21
allow vendor specific lib for saithrift server compile

vincent-xs:
when building a vendor specific libsai.so to link into the saithrift server, we might need extra libraries to finalize the saiserver executable, e.g. dependency on google protocol buffer specific library.
by setting a new environment variable, the dash-pipeline makefile will make sure that
1/ this variable is passed via docker to the saithrift server docker run
2/ the saithrift server makefile uses this env variable to expand the linker compile flags.

example usage from thirdparty makefile

SAIRPC_VENDOR_EXTRA_LIBS="/dash/dash-pipeline/SAI/lib/libprotobuf.a"
thirdparty-saithrift-server: thirdparty-libsai
	@echo "Build third-party saithrift-server"
	@echo "   Expects libsai.so under $(DASHDIR)/dash-pipeline/SAI/lib"
	SAIRPC_VENDOR_EXTRA_LIBS=$(SAIRPC_VENDOR_EXTRA_LIBS) $(MAKE) -C $(DASHDIR)/dash-pipeline saithrift-server

misc:
modify the dash-pipeline makefile to guarantee PWD is the makefile location, no matter where it's called from.

### DIFF
--- a/dash-pipeline/README-dash-workflows.md
+++ b/dash-pipeline/README-dash-workflows.md
@@ -253,7 +253,7 @@ Its value will be added to the baseline SAIRPC_EXTRA_LIBS as defined in the sait
 
 Since the saithrift server is built within a docker container (and the parent repository is mounted as /dash), any of the extra libraries needed will need to be copied over under the parent repository, and the paths to those libraries will need to be relative to the docker mount point.
 
-In the example below, libprotobuf.a is a new external dependancy to the vendor specific libsai.so and has been copied over under the parent repository (in our case, dash-pipeline/SAI/lib).
+In the example below, libprotobuf.a is a new external dependency to the vendor specific libsai.so and has been copied over under the parent repository (in our case, dash-pipeline/SAI/lib).
 We use the provided Makefile.3rdpty as an entry point into the DASH makefiles.
 
 ```
--- a/dash-pipeline/README-dash-workflows.md
+++ b/dash-pipeline/README-dash-workflows.md
@@ -247,6 +247,23 @@ This builds a saithrift-server daemon, which is linked to the `libsai` library a
 ```
 make saithrift-server
 ```
+In the case a vendor integrates its own `libsai` library into the saithrift server, the libsai might have new external dependencies (such as google protocol buffer) thus requiring for vendor specific libraries or linker options to be passed down to the saiserver linker.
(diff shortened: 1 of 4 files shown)


chrispsommers (changes requested):
Code looks good and it passed the CI pipeline. Could you add a little content to https://github.com/vincent-xs/DASH/blob/vincent-xs/vendor_lib/dash-pipeline/README-dash-workflows.md#build-saithrift-server to show this usage? Thanks.

dash-pipeline/README-dash-workflows.md
```diff
@@ -247,6 +247,23 @@ This builds a saithrift-server daemon, which is linked to the `libsai` library a
...
 make saithrift-server
 ```
+In the case a vendor integrates its own `libsai` library into the saithrift server, the libsai might have new external dependencies (such as google protocol buffer) thus requiring for vendor specific libraries or linker options to be passed down to the saiserver linker.
+An environment variable (SAIRPC_VENDOR_EXTRA_LIBS) can be specified when invoking the saithrift server build command to provide path to new libraries and/or new linker options.
+Its value will be added to the baseline SAIRPC_EXTRA_LIBS as defined in the saithrift makefile.
+
+Since the saithrift server is built within a docker container (and the parent repository is mounted as /dash), any of the extra libraries needed will need to be copied over under the parent repository, and the paths to those libraries will need to be relative to the docker mount point.
+
+In the example below, libprotobuf.a is a new external dependancy to the vendor specific libsai.so and has been copied over under the parent repository (in our case, dash-pipeline/SAI/lib).
+We use the provided Makefile.3rdpaty as an entry point into the DASH makefiles.
```
chrispsommers, line 257: typo: `Makefile.3rdpaty` -> `Makefile.3rdpty`
Otherwise looks really good, thanks.

vincent-xs: fixed

chrispsommers (approved):

chrispsommers:
Hi Vincent, welcome to spellcheck. Please add new buzzwords to DASH/.wordlist.txt and/or fix typos identified in this CI run:
https://github.com/Azure/DASH/actions/runs/3101665582/jobs/5023435111

As an aside, I notice you don't have any CI actions in your fork: https://github.com/vincent-xs/DASH/actions I know for my own development fork every push generates a CI run and I can fix a lot of things before I do a PR.

PR #238, opened 2022-09-28
Remove deny action from direction lookup

marian-pritsak:
Make inbound direction a default of a table
Remove the deny option from SAI API as a result


marian-pritsak:
Changes to SAI API are a part of the other updates done before with the same goal and are submitted to SAI in this PR https://github.com/opencomputeproject/SAI/pull/1612

chrispsommers:
LGTM, CI passed too.

chrispsommers (approved):
LGTM, CI passed too.

vmytnykx (approved):

PR #240, opened 2022-09-29
Make SAI attrs mandatory for drop-on-miss tables

marian-pritsak:
If a table references another table (ENI->VNET_ID) where VNET table drops on miss, the SAI attribute VNET_ID in the ENI will be mandatory on create.

Omitting this attribute will no longer be allowed.

### DIFF
--- a/dash-pipeline/SAI/sai_api_gen.py
+++ b/dash-pipeline/SAI/sai_api_gen.py
@@ -233,11 +233,21 @@ def generate_sai_apis(program, ignore_tables):
                         break
 
         param_names = []
+        deny_on_miss = False
         for action in table[ACTION_REFS_TAG]:
             action_id = action["id"]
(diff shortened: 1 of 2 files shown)


marian-pritsak:
Requires https://github.com/opencomputeproject/SAI/pull/1616 and also ported to dash branch.

dash-pipeline/SAI/sai_api_gen.py
```diff
@@ -403,6 +422,8 @@ def write_sai_files(sai_api):
                 for table_name in all_table_names:
                     if table_ref.endswith(table_name):
                         param[OBJECT_NAME_TAG] = table_name
+                        if is_table_deny_on_miss(sai_apis, table_name):
```
chrispsommers, line 425: I would prefer this logic not be buried in the API generator. Can we use a P4 annotation instead?

chrispsommers:
@marian-pritsak is this obsolete?

PR #244, opened 2022-10-03
Address comments from HA Design review

SanjayTh:
Comments from Review including
- Adding SAI Examples
- Init confitguation for Role etc.
- Adds preemtion
- Updated SM
- More details on flow synchronization including what is carried in metadata.
- More details on switchover, flow reconcilation
- etc.


documentation/high-avail/AMD-Pensando_HA_Proposal.md
```diff
@@ -157,50 +222,141 @@ In this state the node is waiting for the Datapath to signal completion of state
...
+
+    /**
+     * @brief L4 Information (TCP/UDP/ICMP)
+     */
+    sai_dash_ha_flow_l4_info_t l4_info;
+
+    /**
+     * @brief policy results metadata
+     */
+     sai_u8_list_t metadata;
```
lguohan, line 322: what is the format of this metadata?

KrisNey-MSFT:
Please try to fix spellcheck :)  merging PR

documentation/high-avail/AMD-Pensando_HA_Proposal.md
```diff
@@ -209,140 +365,221 @@ typedef struct _sai_dash_flow_icmp_info_t {
...
 
     /**
-     * @brief L4 Information (TCP/UDP/ICMP)
+     * @brief Peer's IP address
+     *
+     * @type sai_ip_address_t
+     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
      */
-    sai_dash_flow_l4_info_t l4_info;
+    SAI_DASH_HA_SESSION_ATTR_PEER_IP,
```
lguohan, line 414: need source IP for HA_SESSION.

KrisNey-MSFT: Discussed in HA call, Sanjay to update

documentation/high-avail/AMD-Pensando_HA_Proposal.md
```diff
@@ -209,140 +365,221 @@ typedef struct _sai_dash_flow_icmp_info_t {
...
-/**
- * @brief Attributes ID for get_peer_capabilities
- */
-typedef enum  _sai_get_peer_capabilities_attr_t {
     /**
      * @brief HB Interval
      * @type sai_uint16_t
      */
-    SAI_DASH_GET_PEER_CAPABILITIES_ATTR_HB_INTERVAL,
+    SAI_DASH_HA_SESSION_ATTR_HB_INTERVAL,
```
lguohan, line 429: can you add time granularity for clarification?

documentation/high-avail/AMD-Pensando_HA_Proposal.md
```diff
@@ -209,140 +365,221 @@ typedef struct _sai_dash_flow_icmp_info_t {
...
-    SAI_DASH_GET_PEER_CAPABILITIES_ATTR_HB_MISS_COUNT,
+    SAI_DASH_HA_SESSION_ATTR_HB_MISS_COUNT,
 
     /**
-     * @brief Capabilities
+     * @brief Named pipe for bi-directional control stream
      * @type sai_uint8_list_t
      */
-    SAI_DASH_GET_PEER_CAPABILITIES_ATTR_CAPABILITIES,
+    SAI_DASH_HA_REGISTER_CP_CHANNEL_ATTR_NAMED_PIPE,
```
lguohan, line 441: named pipe is use for get flow_sync messages

documentation/high-avail/AMD-Pensando_HA_Proposal.md
```diff
@@ -157,50 +222,141 @@ In this state the node is waiting for the Datapath to signal completion of state
...
+In this state the DPU is notified to switchover to standalone. The Primary (this node's peer) triggers this once the network convergence timeout has expired on the primary side. The Datapath switches to an intermediate state PRE_STANDALONE. In this state, datapath still accepts flow sync acknowledgments that were in flight from the peer. This state is needed for a duration of time so as to allow for draining all in flight messages.
+
+**Wait Switch Standalone**
+Once the DPU drain timer expires the DPU notifies switchover done which takes the state machine to exit this state and move to full standalone primary. A message is sent to the old primary to shutdown before exiting this state.
 
-This is the terminal state for the node when the node is configured as the primary. In this state the local datapath is forwarding traffic actively and synching state to the peer.
 
 ### SAI Definitions
 
 The SAI API calls necessary for communication between the SONIC stack and the SAI-DASH SDK are defined below.
```
KrisNey-MSFT, line 247: Could we add a section re: callback mechanism?  Needs to be defined so SyncD can register (ReDis is not used).
Should we use 2 different channels? Named pipe for data, another for all control messages?
We could do this, Sanjay and Guohan to look at examples.

documentation/high-avail/AMD-Pensando_HA_Proposal.md
```diff
@@ -50,16 +53,41 @@ Each ENI configured on the DPU is associated with a DP-VIP by the controller. In
 
 ### Datapath Heartbeat
 
-Each DPU sends heartbeat messages at a configured interval to its peer. When a peer loses a set number of heartbeats it declares the peer unreachable and a switchover is initiated. The interval between heartbeats and the number of missed heartbeats are configurable. The Heartbeats can be aggressive hence it is left to the DPU to perform the heartbeat.
+Each DPU sends heartbeat messages at a configured interval to its peer. When a peer loses a set number of heartbeats it declares the peer unreachable and a switchover is initiated. The interval between heartbeats and the number of missed heartbeats are configurable. The Heartbeats can be aggressive hence it is left to the DPU to perform the heartbeat. The Heartbeats can be aggressive hence it is left to the DPU to perform the heartbeat.
```
KrisNey-MSFT, line 56: Should we define the protocol used (UDP, BFD, other)?  Source / Dest ports?  If BFD performs multiphop probing, we could use this.  Need to think more on this...

reshmaintel: Also to check if multipath probing is implemented in SONiC BFD implementation. If HA inline sync utilizes any available path between active and standby (versus fixed path) then multipath probing would be required.

reshmaintel:
Retransmission for inline sync is not very clear in the document. The design is that the retransmission from the source will in case of a link failure in the path. If the standby is not in the path as the source, active and end host, then retransmission will not cover the packet drop between active and standby link. If same packet is expected to be sent to standby and back to active and then forwarded to destination, this will cover all the cases and should be explicitly mentioned in the document

PR #269, opened 2022-10-28
Change list match to prefix list for IPs

marian-pritsak:

### DIFF
--- a/dash-pipeline/SAI/sai_api_gen.py
+++ b/dash-pipeline/SAI/sai_api_gen.py
@@ -68,13 +68,13 @@ def get_sai_list_type(key_size, key_header, key_field):
     elif key_size <= 16:
         return 'sai_u16_list_t', "u16list"
     elif key_size == 32 and ('addr' in key_field or 'ip' in key_header):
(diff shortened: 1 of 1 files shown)


marian-pritsak:
The SAI API change https://github.com/opencomputeproject/SAI/pull/1630

dash-pipeline/SAI/sai_api_gen.py
```diff
@@ -68,13 +68,13 @@ def get_sai_list_type(key_size, key_header, key_field):
     elif key_size <= 16:
         return 'sai_u16_list_t', "u16list"
     elif key_size == 32 and ('addr' in key_field or 'ip' in key_header):
-        return 'sai_ip_address_list_t', "ipaddrlist"
+        return 'sai_ip_prefix_list_t', "ipprefixlist"
```
chrispsommers, line 71: Long-term, you might want to parse the P4Info to get a custom list-type annotation and pass that to this function as the preferred way to generate it, and only use this heuristic approach as a fallback.

chrispsommers (approved):
LGTM but see comment about long-term approach.

PR #273, opened 2022-11-03
[hld]: Support for multiple routing type actions

theasianpianist:
Some routing types require multiple actions to be performed in order. This PR updates the `DASH_ROUTING_TYPE_TABLE` schema to allow for this by adding a second schema for the table. The existing schema will be used for all routing types requiring only a single action. Routing types requiring multiple actions will use the new schema.

Signed-off-by: Lawrence Lee <lawlee@microsoft.com>

### DIFF
--- a/documentation/general/dash-sonic-hld.md
+++ b/documentation/general/dash-sonic-hld.md
@@ -27,17 +27,18 @@
 
 ###### Revision
 
-| Rev |     Date    |       Author          | Change Description                  |
-|:---:|:-----------:|:---------------------:|:------------------------------------|
-| 0.1 | 02/01/2022  |     Prince Sunny      | Initial version                     |
-| 0.2 | 03/09/2022  |     Prince Sunny      | Packet Flows/DB Objects             |
-| 0.3 | 05/24/2022  |      Oleksandr        | Memory Footprints                   |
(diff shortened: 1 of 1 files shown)


documentation/general/dash-sonic-hld.md
```diff
@@ -273,20 +274,34 @@ dst_port                 = list of range of destination ports ',' separated
...
+action_type              = action_type can be {maprouting, direct, staticencap, appliance, 4to6, mapdecap, decap, drop}
+encap_type               = encap type depends on the action_type - {vxlan, nvgre}
+vni                      = vni value associated with the corresponding action. Applicable if encap_type is specified. 
 ```
 
+The second schema is used when multiple, ordered actions are associated with a routing type:
 ```
-key                      = DASH_ROUTING_TYPE_TABLE:routing_type; routing type can be {direct, vnet, vnet_direct, vnet_encap, appliance, privatelink, privatelinknsg, servicetunnel}; actions can be a list of action_types
+DASH_ROUTING_TYPE_TABLE:{{routing_type}}:{{action_name}}
+        "action_type": {{action_type}}
```
theasianpianist, line 296: Further changes will be needed here since we cannot guarantee that all actions for a routing type will be programmed at the same time - there may be a delay between them which need to be handled before we push the updates to SAI

PR #274, opened 2022-11-03
Replace chrissommers dockerhub with sonicdash.azurecr.io (missed in PR225)

chrispsommers:
Two Dockerfiles used some base images in `FROM` clauses and they hadn't gotten changed to use ACR, this fixes that omission. No functional changes.


chrispsommers:
Note, CI scripts fail in the "publish" phase ([Example](https://github.com/Azure/DASH/actions/runs/3387849959/jobs/5629121728#step:4:8)) because:
- changes to Dockerfiles trigger a rebuild and publishing of Docker images
- The PR context is from a fork, which lacks permissions to publish to ACR

Upon merge accept, this change should succeed in the CI because the context will be the main repo, not a fork.

KrisNey-MSFT:
Per Community discussion today

PR #280, opened 2022-11-10
Remove RIF SMAC attribute from HLD

vmytnykx:
Based on BMv2 discussion, the SRC mac of egress packet should be set based on SMAC of switch, not RIF. So, looks like this RIF SAI_ROUTER_INTERFACE_ATTR_SRC_MAC_ADDRESS attribute is redundant in the doc, thus removing it. The SAI_SWITCH_ATTR_SRC_MAC_ADDRESS should used instead.


prsunny:
In general, the switch mac shall be same as RIF mac, but for packet forwarding, it is expected to use the RIF mac as src mac. So please keep the attribute.

PR #281, opened 2022-11-10
Add bmv2 libsai switch & port attributes

### LINKED ISSUE #235: sai_get_switch-attribute() returns 0 when requesting number_of_active_ports
The issue was discovered when running tests from PR #231.

**This issue blocks all SAI PTF tests run.**
self.portX objects cannot be initialized. Standard SAI PTF fails on SaiHelperBase.get_active_port_list() method.
And adapted DASH PTF tests are skipped in SaiHelperSimplified.__getattr__ because of self.port0 does not exist.

Log:
```
(Pdb++) sai_thrift_get_switch_attribute(self.client, number_of_active_ports=True)
{'SAI_SWITCH_ATTR_NUMBER_OF_ACTIVE_PORTS': 0, 'number_of_active_ports': 0}
```

chrispsommers:
Fixes #235. Added fake switch & port attributes to bmv2 libsai in order to satisfy PyTest and PTF base classes. Use `SaiHelperSimplified` as PTF base class to replace workaround use of `ThriftInterfaceDataPlane` base class.

### DIFF
--- a/dash-pipeline/SAI/templates/utils.cpp.j2
+++ b/dash-pipeline/SAI/templates/utils.cpp.j2
@@ -217,12 +217,78 @@ sai_status_t sai_create_switch_dummy(
     return SAI_STATUS_SUCCESS;
 }
 
-sai_status_t sai_get_switch_attribute_dummy(
+#define DASH_BMV2_NUM_PORTS 2 
+#define DASH_BMV2_CPU_QOS_NUMBER_OF_QUEUES 0
+#define DASH_BMV2_DEFAULT_CPU_PORT_ID (sai_object_id_t)((SAI_OBJECT_TYPE_PORT<<48)+1)
+#define DASH_BMV2_DEFAULT_VLAN_ID (sai_object_id_t)((SAI_OBJECT_TYPE_VLAN<<48)+1)
+#define DASH_BMV2_DEFAULT_VRF_ID (sai_object_id_t)((SAI_OBJECT_TYPE_VIRTUAL_ROUTER<<48)+1)
+#define DASH_BMV2_DEFAULT_1Q_BRIDGE_ID (sai_object_id_t)((SAI_OBJECT_TYPE_BRIDGE<<48)+1)
+#define DASH_BMV2_DEFAULT_1Q_BRIDGE_ID (sai_object_id_t)((SAI_OBJECT_TYPE_BRIDGE<<48)+1)
+
+sai_object_id_t port_list[DASH_BMV2_NUM_PORTS] = {
+    (sai_object_id_t)((SAI_OBJECT_TYPE_PORT<<48) + 1),
+    (sai_object_id_t)((SAI_OBJECT_TYPE_PORT<<48) + 2)
+};
+
+sai_status_t sai_get_switch_attribute(
         _In_ sai_object_id_t switch_id,
         _In_ uint32_t attr_count,
         _Inout_ sai_attribute_t *attr_list) {
 
-    fprintf(stderr, "sai_get_switch_attribute_dummy()\n");
+    fprintf(stderr, "sai_get_switch_attribute()\n");
+    int i;
+    sai_attribute_t *attr = attr_list;
+    sai_object_list_t port_obj_list;
+    sai_object_id_t *objlist;
+    for (i = 0; i < attr_count ; i++, attr++) {
+        switch(attr->id) {
+        
+        case SAI_SWITCH_ATTR_NUMBER_OF_ACTIVE_PORTS:
+            attr->value.u32 = DASH_BMV2_NUM_PORTS;
+            fprintf(stderr, "  sai_get_switch_attribute() [%d] attr %d SAI_SWITCH_ATTR_NUMBER_OF_ACTIVE_PORTS = %d\n",
+                    i, attr->id, attr->value.u32);
+            return SAI_STATUS_SUCCESS;
+        
+        case SAI_SWITCH_ATTR_PORT_LIST:
+            // make a tmp port list, saiserver will free the memory
+            objlist = (sai_object_id_t *)malloc(sizeof(port_list));
+            memcpy(objlist, port_list, sizeof(port_list));
+            port_obj_list = {
+                .count = DASH_BMV2_NUM_PORTS,
+                .list = objlist
+            };                
+            attr->value.objlist = port_obj_list;
+            fprintf(stderr, "  sai_get_switch_attribute() [%d] attr %d SAI_SWITCH_ATTR_PORT_LIST = [%d objids]\n",
+                    i, attr->id, DASH_BMV2_NUM_PORTS);
+            return SAI_STATUS_SUCCESS;
+        
+        case SAI_SWITCH_ATTR_DEFAULT_VLAN_ID:
+            attr->value.oid = DASH_BMV2_DEFAULT_VLAN_ID;
+            fprintf(stderr, "  sai_get_switch_attribute() [%d] attr %d SAI_SWITCH_ATTR_DEFAULT_VLAN_ID = %lx\n",
+                    i, attr->id, attr->value.oid);
+            return SAI_STATUS_SUCCESS;
+        
+        case SAI_SWITCH_ATTR_DEFAULT_VIRTUAL_ROUTER_ID:
+            attr->value.oid = DASH_BMV2_DEFAULT_VRF_ID;
+            fprintf(stderr, "  sai_get_switch_attribute() [0] attr %d SAI_SWITCH_ATTR_DEFAULT_VIRTUAL_ROUTER_ID = %lx\n", attr->id, attr->value.oid);
+            return SAI_STATUS_SUCCESS;
+        
+        case SAI_SWITCH_ATTR_DEFAULT_1Q_BRIDGE_ID:
+            attr->value.oid = DASH_BMV2_DEFAULT_1Q_BRIDGE_ID;
+            fprintf(stderr, "  sai_get_switch_attribute() [0] attr %d SAI_SWITCH_ATTR_DEFAULT_1Q_BRIDGE_ID = %lx\n", attr->id, attr->value.oid);
+            return SAI_STATUS_SUCCESS;
+
+        case SAI_SWITCH_ATTR_CPU_PORT:
+            attr->value.oid = DASH_BMV2_DEFAULT_CPU_PORT_ID;
+            fprintf(stderr, "  sai_get_switch_attribute() [%d] attr %d DASH_BMV2_DEFAULT_CPU_PORT_ID = %lx\n",
+                    i, attr->id, attr->value.oid);
+            return SAI_STATUS_SUCCESS;
+
+        default:
+            fprintf(stderr, "  sai_get_switch_attribute() [0] attr %d is NOT SUPPORTED - returning SAI_STATUS_SUCCESS anyway\n", attr->id);
+            return SAI_STATUS_SUCCESS;
+        }
+    }
     return SAI_STATUS_SUCCESS;
 }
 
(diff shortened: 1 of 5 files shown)


chrispsommers:
@aputriax please review as well.

dash-pipeline/tests/saithrift/ptf/thrift/test_thrift_session.py
```diff
@@ -1,7 +1,7 @@
 from sai_thrift.sai_headers import *
 from sai_base_test import *
 
-class TestSaiThriftSession(ThriftInterfaceDataPlane):
+class TestSaiThriftSession(SaiHelperSimplified):
     """ Test saithrift client connection only"""
     def setup(self):
```
aputriax, line 6: FYI: `setup()` and `teardown()` are never called in this class.
The method names should be `setUp()` and `tearDown()` - with camelCase.

dash-pipeline/tests/saithrift/pytest/switch/test_saithrift_switch.py
```diff
@@ -6,10 +6,45 @@
...
+        saithrift_client, cpu_port=True)
+    cpu_port = attr['cpu_port']
+    print ("cpu_port = %d" % cpu_port)
+    assert(cpu_port != 0)
+    
+    attr = sai_thrift_get_port_attribute(
+        saithrift_client, cpu_port, qos_number_of_queues=True)
+    qos_number_of_queues = attr['qos_number_of_queues']
+    print ("qos_number_of_queues = %d" % qos_number_of_queues)
+    assert(qos_number_of_queues == 0)
```
aputriax, line 47: Maybe we should check `qos_number_of_queues is not None` so when the queues are implemented this test won't fail?

dash-pipeline/tests/saithrift/ptf/thrift/test_thrift_session.py
chrispsommers: Fixed,thanks!

dash-pipeline/tests/saithrift/pytest/switch/test_saithrift_switch.py
chrispsommers: If and when the queues are implemented we can change the test. I wouldn't expect QoS for CPU port to be implemented soon.

dash-pipeline/tests/saithrift/pytest/switch/test_saithrift_switch.py
```diff
@@ -6,10 +6,46 @@
...
+    assert(number_of_active_ports == attr['port_list'].count)
+    port_list = attr['port_list'].idlist
+    print ("port list = ", port_list)
+    assert(port_list  is not None)
+
+    attr = sai_thrift_get_switch_attribute(
+        saithrift_client, default_vlan_id=True)
+    default_vlan_id = attr['default_vlan_id']
+    print ("default_vlan_id = %d" % default_vlan_id)
+    assert(default_vlan_id  is not None)
```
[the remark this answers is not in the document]
chrispsommers: @aputriax done, thanks!

dash-pipeline/SAI/templates/utils.cpp.j2
```diff
@@ -217,20 +217,86 @@ sai_status_t sai_create_switch_dummy(
...
+        
+        case SAI_SWITCH_ATTR_NUMBER_OF_ACTIVE_PORTS:
+            attr->value.u32 = DASH_BMV2_NUM_PORTS;
+            fprintf(stderr, "  sai_get_switch_attribute() [%d] attr %d SAI_SWITCH_ATTR_NUMBER_OF_ACTIVE_PORTS = %d\n",
+                    i, attr->id, attr->value.u32);
+            return SAI_STATUS_SUCCESS;
+        
+        case SAI_SWITCH_ATTR_PORT_LIST:
+            // make a tmp port list, saiserver will free the memory
+            objlist = (sai_object_id_t *)malloc(sizeof(port_list));
```
vmytnykx, line 254: I think we can use cpp style allocation instead of C malloc

dash-pipeline/SAI/templates/utils.cpp.j2
```diff
@@ -243,6 +309,62 @@ sai_switch_api_t sai_switch_api_impl = {
...
+        default:
+            fprintf(stderr, "  sai_get_port_attribute() [0] attr %d is NOT SUPPORTED - returning SAI_STATUS_SUCCESS anyway\n", attr->id);
+            return SAI_STATUS_SUCCESS;
+        }
+    }
+
+    return SAI_STATUS_SUCCESS;
+}
+
+sai_port_api_t sai_port_api_impl = {
```
vmytnykx, line 337: Make this var static and you don't need = (void *)0 for each of the fields as they will be zero anyway.

vmytnykx (commented):
@chrispsommers thanks for making those changes. In general it looks good with minor comments.

dash-pipeline/SAI/templates/utils.cpp.j2
chrispsommers: Hi @vmytnykx I did make static as you suggested, but left in the explicit zero values as a way to document the available APIs and what is not implemented. (Removing the (void *) cleanup up some warnings though).

dash-pipeline/SAI/templates/utils.cpp.j2
chrispsommers: If you have a specific suggestion I'll consider it, otherwise I'm happy with this since it's plain c-code in any case.

dash-pipeline/SAI/templates/utils.cpp.j2
```diff
@@ -217,20 +217,86 @@ sai_status_t sai_create_switch_dummy(
     return SAI_STATUS_SUCCESS;
 }
 
-sai_status_t sai_get_switch_attribute_dummy(
+#define DASH_BMV2_NUM_PORTS 2 
+#define DASH_BMV2_CPU_QOS_NUMBER_OF_QUEUES 0
+#define DASH_BMV2_DEFAULT_CPU_PORT_ID (sai_object_id_t)((SAI_OBJECT_TYPE_PORT<<48)+1)
```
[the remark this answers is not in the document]
chrispsommers: I was referring to the practice defined [here](https://github.com/opencomputeproject/SAI/blob/master/doc/SAI-Proposal-7-UOID.md#specification) where the object type is encoded in the upper 16 bits of a uint64_t. I defined some macros to make it more explicit and readable. Let me know if there is still any concern.

Also note I changed CPU port to 64 decimal to differentiate from the two dummy traffic ports 1 and 2, plus changed the printing in `tests/saithrift/pytest/switch/test_saithrift_switch.py` to he to make it easier to see the object_id's 16 and 48-bit components. e.g.:
```
sai-thrift connection established with localhost:9092
number_of_active_ports = 2
port list =  ['0x0001000000000001', '0x0001000000000002']
default_vlan_id = 0x0026000000000001
default_virtual_router_id = 0x0003000000000001
cpu_port = 0x0001000000000040
qos_number_of_queues = 0
test_sai_thrift_get_switch_attribute OK
```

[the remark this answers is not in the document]
vmytnykx: I'm ok with this practice. What I mean is just define this 48 number as it is used in different places. This is C good practice.

vmytnykx (approved):

chrispsommers:
Hi @marian-pritsak could you please take a quick look at this? It's needed to unblock some other work, thanks!

lguohan:
@marian-pritsak , can you review this?

dash-pipeline/SAI/templates/utils.cpp.j2
```diff
@@ -217,29 +217,155 @@ sai_status_t sai_create_switch_dummy(
...
+#define DASH_BMV2_DEFAULT_VRF_ID MAKE_OID(SAI_OBJECT_TYPE_VIRTUAL_ROUTER,1)
+#define DASH_BMV2_DEFAULT_1Q_BRIDGE_ID MAKE_OID(SAI_OBJECT_TYPE_BRIDGE,1)
+#define DASH_BMV2_DEFAULT_1Q_BRIDGE_ID MAKE_OID(SAI_OBJECT_TYPE_BRIDGE,1)
+
+sai_object_id_t port_list[DASH_BMV2_NUM_PORTS] = {
+    MAKE_OID(SAI_OBJECT_TYPE_PORT,1),
+    MAKE_OID(SAI_OBJECT_TYPE_PORT,2)
+};
+
+sai_status_t sai_get_switch_attribute(
```
lguohan, line 237: can we move this to saiswitch.cpp?

dash-pipeline/SAI/templates/utils.cpp.j2
```diff
@@ -217,29 +217,155 @@ sai_status_t sai_create_switch_dummy(
     return SAI_STATUS_SUCCESS;
 }
 
-sai_status_t sai_get_switch_attribute_dummy(
+#define DASH_BMV2_NUM_PORTS 2 
+#define DASH_BMV2_CPU_QOS_NUMBER_OF_QUEUES 0
+
+#define OJBTYPE_SHFT 48
```
lguohan, line 223: is this OBJTYPE_SHIFT?

dash-pipeline/SAI/templates/utils.cpp.j2
```diff
@@ -217,29 +217,155 @@ sai_status_t sai_create_switch_dummy(
...
+#define MAKE_OID(_objtype, _objval) (sai_object_id_t)(((sai_object_id_t)_objtype<<OJBTYPE_SHFT)+(sai_object_id_t)_objval)
+
+#define DASH_BMV2_DEFAULT_CPU_PORT_ID MAKE_OID(SAI_OBJECT_TYPE_PORT,64)
+#define DASH_BMV2_DEFAULT_VLAN_ID MAKE_OID(SAI_OBJECT_TYPE_VLAN,1)
+#define DASH_BMV2_DEFAULT_VRF_ID MAKE_OID(SAI_OBJECT_TYPE_VIRTUAL_ROUTER,1)
+#define DASH_BMV2_DEFAULT_1Q_BRIDGE_ID MAKE_OID(SAI_OBJECT_TYPE_BRIDGE,1)
+#define DASH_BMV2_DEFAULT_1Q_BRIDGE_ID MAKE_OID(SAI_OBJECT_TYPE_BRIDGE,1)
```
lguohan, line 230: seems dup of line 229?

dash-pipeline/SAI/templates/utils.cpp.j2
chrispsommers: I'll give it a try,thanks.

dash-pipeline/SAI/templates/utils.cpp.j2
chrispsommers: Yes, a cut and paste error. Thanks.

dash-pipeline/SAI/templates/utils.cpp.j2
chrispsommers: I created a `saifixedapis.cpp.j2` file which contains a mix of API callbacks, not just switch-related. If this grows too large we can split off into more files. I'm trying to strike a balance between modularity and file bloat.

dash-pipeline/SAI/templates/utils.cpp.j2
chrispsommers: Fixed.

dash-pipeline/SAI/templates/utils.cpp.j2
chrispsommers: Used to construct an OID from object type and instance value. See macro `MAKE_OID`

marian-pritsak (approved):
Looks good.

PR #282, opened 2022-11-11
Refactor saichallenger dockerfiles

chrispsommers:
Per review feedback, build sai-challenger images in fork to handle case where images are not in ACR yet. Main repo builds will publish and pull. Also we now build saichallenger-client from a relatively stable base image (saichallenger-client-bldr) plus changeable portion based on DASH P4 code/ilbsai/thrift client code, this speeds things up when saichallenger-client-bldr is already pre-built in a docker registry.

Note, a future PR will use a more intelligent build-vs-pull docker image strategy across all images, to build on demand (and possibly publish) only when an image isn't available in a registry.


chrispsommers:
Merging into staging branch only, it will allow publishing to ACR and CI will pass.

PR #295, opened 2022-12-07
Fixes for sai scale test to use dynamically generated dpu config instead of static.

vkuma82:
DPU generator used to create the dpu config instead of static json file.
Bump the dpugen version to 0.0.5.


chrispsommers (approved):
LGTM

PR #296, opened 2022-12-08
Add DASH outbound metering to Sonic HLD

vijasrin:
- Metering Policy, rule and meter schemas
    - Metering feature description
    - Update DASH outbound packet processing pipeline
    - Update meter examples
    - Adding pipeline diagram for outbound meter


vijasrin:
@prsunny

documentation/general/dash-sonic-hld.md
```diff
@@ -220,6 +243,8 @@ DASH_ENI_TABLE:{{eni}}
     "vnet": {{vnet_name}}
     "pl_sip_encoding": {{string}} (OPTIONAL)
     "pl_underlay_sip": {{ip_addr}} (OPTIONAL)
+    "v4_meter_policy_id": {{string}} (OPTIONAL)
+    "v6_meter_policy_id": {{string}} (OPTIONAL)
```
lguohan, line 247: why these are string instead of integer?

documentation/general/dash-sonic-hld.md
```diff
@@ -386,13 +417,59 @@ key                      = DASH_VNET_MAPPING_TABLE:vnet:ip_address ; CA-PA mappi
...
+priority                 = INT32 value  ; priority of the rule: lower the value, higher the priority
+ip_prefix                = ip_prefix    ; ip prefix for matching
+metering_bucket          = bucket_id    ; metering bucket id
+```
+
+
+```
+DASH_METER:{{eni}}:{{bucket_id}}
+    “bucket_metadata”: {{string}} (OPTIONAL)
+    "tx_counter": {{bytes}}
```
lguohan, line 460: maybe we should rename this as inbound and outbound counter

documentation/general/dash-sonic-hld.md
```diff
@@ -386,13 +417,59 @@ key                      = DASH_VNET_MAPPING_TABLE:vnet:ip_address ; CA-PA mappi
...
+key                      = DASH_METER_POLICY:meter_policy_id ; policy_id as key
+; field                  = value 
+ip_version               = IP version (IPv4/IPv6)
+```
+
+
+```
+DASH_METER_RULE:{{meter_policy_id}}:{{rule_num}}
+    "priority": {{priority}}
+    "ip_prefix":{{ip_prefix}}
```
lguohan, line 444: which ip address to match in case of pl/st case?

documentation/general/dash-sonic-hld.md
```diff
@@ -386,13 +417,59 @@ key                      = DASH_VNET_MAPPING_TABLE:vnet:ip_address ; CA-PA mappi
...
+key                      = DASH_METER_RULE:meter_policy_id:rule_num ; unique rule num within the policy.
+; field                  = value 
+priority                 = INT32 value  ; priority of the rule: lower the value, higher the priority
+ip_prefix                = ip_prefix    ; ip prefix for matching
+metering_bucket          = bucket_id    ; metering bucket id
+```
+
+
+```
+DASH_METER:{{eni}}:{{bucket_id}}
```
mhanif, line 458: I take it that the DASH_METER is essentially the "bucket" which we refer to in this document.  Also, it has a "bucket_id".  Why do we need "ENI" as a key here?  Since, as you mentioned earlier, the bucket IDs are globally unique across all VNETs in a region, I don't believe we need the ENI as a key here.  Also, bucket IDs are assigned to policy meters and policy meters are subsequently attached to ENIs, the usage of ENI here doesn't make sense.  Am I missing something here?  Thanks!

documentation/general/dash-sonic-hld.md
vijasrin: these are meter policy UUIDs, similar to eni-id

documentation/general/dash-sonic-hld.md
vijasrin: IP address used for lookup will be the packet CA Destination IP address. Although for PL case, the bucket-id is driven by the mapping and for ST the bucket-id is driven by the route

documentation/general/dash-sonic-hld.md
vijasrin: inbound/outbound terminology is preiously used in the context of the first (initiator) packet for the connection. In this case, the tx/rx counter can be for the same outbound (VM initiated) connection

documentation/general/dash-sonic-hld.md
vijasrin: The billing is on a per ENI basis, so the global metering bucket id needs to be associated with each ENI where metering is required. Also metering buckets are associated with mappings as well which is per VNET

KrisNey-MSFT:
This is mainly on App_DB schema and will go independently.

documentation/general/dash-sonic-hld.md
```diff
@@ -370,7 +399,8 @@ DASH_ROUTE_TABLE:{{eni}}:{{prefix}}
     "overlay_dip":{{ip_address}} (OPTIONAL)
     "underlay_sip":{{ip_address}} (OPTIONAL)
     "underlay_dip":{{ip_address}} (OPTIONAL)
-    "metering_bucket": {{bucket_id}} (OPTIONAL) 
+    "metering_policy_en": {{bool}}
```
prsunny, line 402: this must be marked OPTIONAL

vijasrin: done

mhanif (commented):
Suggestion for a minor name change

documentation/general/dash-sonic-hld.md
```diff
@@ -439,13 +474,59 @@ key                      = DASH_VNET_MAPPING_TABLE:vnet:ip_address ; CA-PA mappi
...
+key                      = DASH_METER_RULE:meter_policy_id:rule_num ; unique rule num within the policy.
+; field                  = value 
+priority                 = INT32 value  ; priority of the rule: lower the value, higher the priority
+ip_prefix                = ip_prefix    ; ip prefix for matching
+metering_class           = class_id     ; metering class-id
+```
+
+
+```
+DASH_METER:{{eni}}:{{class_id}}
```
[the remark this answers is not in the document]
vijasrin: okay. done

KrisNey-MSFT:
Hanif & Guohan please approve if you have time

mhanif (approved):
LGTM

prsunny:
Merging the PR to proceed with implementation. Further comments can be addressed in a future PR.

PR #304, opened 2022-12-14
Adding underlay BGP testcase

desaimg1:
This test covers underlay bgp established verification tested with snappi


mgheorghe:
TODO: add this in sonic-mgmt

PR #305, opened 2022-12-16
Updated run-functional-tests.sh

aputriax:
Added options:
- SAI Server IP address
- Paths to port map and port config files
- Underlay/Overlay tests selection
- Hardware/BMv2 selection
- Interfaces specifing

Signed-off-by: Yuriy Harhas <yuriyx.harhas@intel.com>

### DIFF
--- a/dash-pipeline/Makefile
+++ b/dash-pipeline/Makefile
@@ -429,7 +429,7 @@ run-saithrift-ptftests: disable-ipv6
 	$(DOCKER_RUN_SAITHRIFT_CLIENT) \
 	-w /tests/ \
 	$(DOCKER_SAITHRIFT_CLIENT_IMG) \
-	./run-functional-tests.sh
+	./functional/ptf/run-tests.sh
 
 # Run "dev" PTF tests mounted from host
 # Note - if add ixia-c usage to PTF tests, add dependency deploy-ixiac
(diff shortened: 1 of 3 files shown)


chrispsommers:
I suggest you consider to rename/move this script to `test/test-cases/functional/ptf/run-tests.sh` and the test path would nominally be `.` I am contemplating doing the same for saic tests under `functional/` and `scale/`

aputriax:
> I suggest you consider to rename/move this script to `test/test-cases/functional/ptf/run-tests.sh` and the test path would nominally be `.` I am contemplating doing the same for saic tests under `functional/` and `scale/`

I can do this, but than I'll need to update Makefile as well. If it is ok, I'll update this PR.

chrispsommers:
> > I suggest you consider to rename/move this script to `test/test-cases/functional/ptf/run-tests.sh` and the test path would nominally be `.` I am contemplating doing the same for saic tests under `functional/` and `scale/`
> 
> I can do this, but than I'll need to update Makefile as well. If it is ok, I'll update this PR.

Updating the Makefile is fine by me, thanks.

chrispsommers (approved):
LGTM

chrispsommers (approved):
LGTM

PR #310, opened 2023-01-03
Add ADDR_FAMILY attr generation

marian-pritsak:
Generate ADDR_FAMILY attribute to object types
that have IP address as part of it's key.

This is a read-only attribute used by resource
availability API.


marian-pritsak:
Generates attributes like:
```
    /**
     * @brief IP address family for resource accounting
     *
     * @type sai_ip_addr_family_t
     * @flags READ_ONLY
     * @isresourcetype true
     */
    SAI_INBOUND_ROUTING_ENTRY_ATTR_IP_ADDR_FAMILY,
```

PR #313, opened 2023-01-11
[Sonic HLD] Metering requirements

prsunny:


documentation/general/dash-sonic-hld.md
```diff
@@ -103,7 +104,34 @@ Following are the minimal scaling requirements
 | CA-PA Mappings           | 10M                      |
 | Active Connections/ENI   | 1M (Bidirectional TCP or UDP)       |
 
-## 1.5 Design Considerations
+## 1.5 Metering requirements
+Metering is essential for billing the customers and below are the high-level requirements
+- Billing shall be at per ENI level and shall be able to query metering packet bytes per ENI
+- All buckets must be UINT64 size and start from value 0 and shall be counting number of bytes
```
KrisNey-MSFT, line 110:
```suggestion
- All buckets must be UINT64 size and start from value 0 and shall be counting number of bytes.  A bucket contains 2 counters; 1 inbound and 1 outbound.  

```

prsunny: Addressed

documentation/general/dash-sonic-hld.md
```diff
@@ -102,8 +103,38 @@ Following are the minimal scaling requirements
 | ACLs per ENI             | 6x10K SRC/DST ports      |
 | CA-PA Mappings           | 10M                      |
 | Active Connections/ENI   | 1M (Bidirectional TCP or UDP)       |
-
-## 1.5 Design Considerations
+| Metering Buckets per ENI | 4000                     |
+
+## 1.5 Metering requirements
+Metering is essential for billing the customers and below are the high-level requirements. Metering/Bucket in this context is related to packet counting for billing purposes and not related to traffic policer or shaping.
```
mhanif, line 109: Instead of "packet counting", we should be "byte counting" while "packet counting" is optional

documentation/general/dash-sonic-hld.md
```diff
@@ -102,8 +103,38 @@ Following are the minimal scaling requirements
...
+| Metering Buckets per ENI | 4000                     |
+
+## 1.5 Metering requirements
+Metering is essential for billing the customers and below are the high-level requirements. Metering/Bucket in this context is related to packet counting for billing purposes and not related to traffic policer or shaping. 
+- Billing shall be at per ENI level and shall be able to query metering packet bytes per ENI
+- All metering buckets must be UINT64 size and start from value 0 and shall be counting number of bytes. A bucket contains 2 counters; 1 inbound (Rx) and 1 outbound (Tx).  
+- Implementation must support metering at the following levels:
+	- Policy based metering. - E.g. For specific destinations (prefix) that must be billed separately, say action_type 'direct'
+	- Route table based metering - E.g. For Vnet peering cases.
+	- Mapping table based metering - E.g For specific destinations within mapping table that must be billed separately
```
mhanif, line 115: minor "the" mapping table

documentation/general/dash-sonic-hld.md
```diff
@@ -102,8 +103,38 @@ Following are the minimal scaling requirements
...
+- Policy in the metering context refers to metering policy associated to Route tables. This is not related to ACL policy or any ACL packet counters.  
+- If packet flow hits multiple metering buckets, order of priority shall be **Policy->Route->Mapping**
+- User shall be able to override the precedence between Routing and Mapping buckets by setting an _override_ flag. When policy is enabled for a route, it takes higher precedence than routing and mapping metering bucket. 
+- Implementation shall aggregate the counters on an "_ENI+Metering Bucket_" combination for billing:
+	- 	All traffic from an ENI to a Peered VNET
+	- 	All traffic from an ENI to a Private Link destination
+	- 	All traffic from an ENI to an internet destination
+	- 	All traffic from an ENI to cross-AZ destination (within the same VNET)
+	- 	All traffic from an ENI to cross-region destination (towards the peer VNET)
+	- 	All outbound metered traffic from an ENI
```
mhanif, line 125: How is this "outbound metered traffic" different from the everything listed before it?  Should this be "All other" outbound metered traffic....."?  When you say implementation, which "implementation"?  Is this "HW pipeline", "SW on the DPU", or "the SDN controller"?  Also, you could have cross-AZ peer VNET traffic as well, would you not want to count separately as well?  I guess I don't understand why you are listing these outbound traffic categories.  Is it that these are the ones you want to bill separately?

documentation/general/dash-sonic-hld.md
```diff
@@ -102,8 +103,38 @@ Following are the minimal scaling requirements
...
+	- 	All inbound metered traffic towards an ENI
+- Customer is billed based on number of bytes sent/received separately. A distinct counter must be supported for outbound vs inbound traffic of each category.
+- For outbound flow and associated metering bucket, created as part of VM initiated traffic, the metering bucket shall account for outbound (Tx) bytes. Based on this outbound flow, pipeline shall also create a unified inbound flow. The same metering bucket shall account for the inbound (Rx) bytes for the return traffic to VM that matches this flow.
```
mhanif, line 128: From scale point of view, there are going be 5M connection and 10M flows (if we count inbound and outbound flows separately).  Let's say we support 5M combined unified flows and if the scale required is 4K buckets per ENI, for a total of 64 ENIs, we will have 256K buckets per DPU.  Considering we are sharing these buckets across ENIs and across flows, we will be limited to how much we can count separately for each of the categories you have listed above?  Is this not true?

documentation/general/dash-sonic-hld.md
prsunny: All outbound metered traffic means query on `ENI + * `-> Aggregated counter from all metering buckets associated with ENI. For cross-AZ peer VNET, requirement is to bill as peer VNET. But this is made generic in the modeling that if it requires to be billed differently, it is still possible. Yes, those are the cases that must be billed separately.
I've updated the meaning for "Implementation" in the first usage.

documentation/general/dash-sonic-hld.md
prsunny: Yes, it is true. But note that many connections may share same bucket id. For example, if there are 100's of mapping for a cross-Az, it will all share the same bucket. However, it is possible that in future we may have to increase the limit from 4k per ENI.

documentation/general/dash-sonic-hld.md
```diff
@@ -102,8 +103,38 @@ Following are the minimal scaling requirements
...
+- User shall be able to override the precedence between Routing and Mapping buckets by setting an _override_ flag. When policy is enabled for a route, it takes higher precedence than routing and mapping metering bucket. 
+- Implementation shall aggregate the counters on an "_ENI+Metering Bucket_" combination for billing:
+	- 	All traffic from an ENI to a Peered VNET
+	- 	All traffic from an ENI to a Private Link destination
+	- 	All traffic from an ENI to an internet destination
+	- 	All traffic from an ENI to cross-AZ destination (within the same VNET)
+	- 	All traffic from an ENI to cross-region destination (towards the peer VNET)
+	- 	All outbound metered traffic from an ENI
+	- 	All inbound metered traffic towards an ENI
+- Customer is billed based on number of bytes sent/received separately. A distinct counter must be supported for outbound vs inbound traffic of each category.
```
vijasrin, line 127: fix terminology: replace inbound/outbound with TX/RX ? since inbound/outbound refers to the first (initiator) packet for the flow

jaepark08 (approved):

documentation/general/dash-sonic-hld.md
prsunny: @vijasrin , i cannot commit to this forked branch anymore after migration. Since its a minor fix, i'll update it in one of the future commits.

KrisNey-MSFT (approved):

PR #316, opened 2023-01-12
Test: Update README.md

xumia:


KrisNey-MSFT:
@xumia can you help us with the new Easy CLA check that was added (PRs are now not passing the check) please?

xumia:
@KrisNey-MSFT , we may need to ask the LinuxFundation to enable the check.
PR author may need to run /EasyCLA, works for my PR.

PR #321, opened 2023-01-18
Change repo links to from Azure/DASH to sonic-net/DASH

chrispsommers:
The following script was used. Empirically, I determined that only `.py, .svg` and `.md` files should be changed. You can also find the `Azure/DASH` string inside `.git` log files but those shouldn't be modified!
```
grep -rnl --include=*.py  --include=*.md --include=*.svg 'github.com/Azure/DASH'|xargs sed -i -e  's.Azure/DASH.sonic-net/DASH.g'
```


yusefMS06 (approved):
Looks good

KrisNey-MSFT:
Thank you @chrispsommers !

PR #324, opened 2023-01-29
Add version of P4 reference code that compiles with p4c-dpdk compiler

jfingerh:
using the Portable NIC Architecture

### DIFF
--- a/dash-pipeline/dpdk-pna/README.md
+++ b/dash-pipeline/dpdk-pna/README.md
@@ -18,7 +18,7 @@ The good news is that with only a few changes in the code, it
 compiles.
 
 There are several warning messages output by p4c-dpdk that are
-currently normal and exepcted.  These are warnings about the DPDK
+currently normal and expected.  These are warnings about the DPDK
 implementation of this P4 program that causes extra data copies during
 packet processing, with a minor effect on the packet processing
 performance.
--- a/dash-pipeline/dpdk-pna/README.md
+++ b/dash-pipeline/dpdk-pna/README.md
@@ -1,0 +1,80 @@
+# Introduction
+
+At the 2023-Jan-26 DASH Behavioral Model WG meeting, I (Andy
+Fingerhut) volunteered to attempt to compile either the unmodified
+DASH P4 reference code in this directory
+
++ https://github.com/sonic-net/DASH/tree/main/dash-pipeline/bmv2
+
+using the open source p4c-dpdk compiler, or to make as few changes as
+possible to it in order for it to compile.  The goal is to determine
+whether there are any features missing in the p4c-dpdk compiler in
+order to compile this P4 code.
+
+
+# Results as of 2023-Jan-28
+
+The good news is that with only a few changes in the code, it
+compiles.
+
+There are several warning messages output by p4c-dpdk that are
+currently normal and exepcted.  These are warnings about the DPDK
+implementation of this P4 program that causes extra data copies during
+packet processing, with a minor effect on the packet processing
+performance.
+
+```
+$ make
+mkdir -p out
+p4c-dpdk -DPNA_CONNTRACK --dump ./out --pp ./out/dash_pipeline.pp.p4 -o ./out/dash_pipeline.spec --arch pna --bf-rt-schema ./out/dash_pipeline.p4.bfrt.json --p4runtime-files ./out/dash_pipeline.p4.p4info.txt dash_pipeline.p4
+[--Wwarn=mismatch] warning: Mismatched header/metadata struct for key elements in table outbound_ConntrackOut_conntrackOut. Copying all match fields to metadata
+[--Wwarn=mismatch] warning: Mismatched header/metadata struct for key elements in table outbound_ConntrackIn_conntrackIn. Copying all match fields to metadata
+[--Wwarn=mismatch] warning: Mismatched header/metadata struct for key elements in table outbound_outbound_ca_to_pa|dash_outbound_ca_to_pa. Copying all match fields to metadata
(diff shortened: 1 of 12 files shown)


jfingerh:
This PR is mostly a copy and paste of the P4 files in the dash-pipeline/bmv2 directory into a new dash-pipeline/dpdk-pna directory, with relatively small changes required to compile the code using the open source p4c-dpdk compiler.

See the README.md file added in this PR for more details.

dash-pipeline/dpdk-pna/README.md
```diff
@@ -0,0 +1,80 @@
...
+order to compile this P4 code.
+
+
+# Results as of 2023-Jan-28
+
+The good news is that with only a few changes in the code, it
+compiles.
+
+There are several warning messages output by p4c-dpdk that are
+currently normal and exepcted.  These are warnings about the DPDK
```
chrispsommers, line 21: spelling - please check spellcheck results in https://github.com/sonic-net/DASH/actions/runs/4039001294/jobs/6943385847 and either correct errors, or add terms to `.wordlist.txt`

chrispsommers (changes requested):
Andy, great work! Thanks a lot for doing this. It's encouraging that you were able to get something to compile without massive changes. We can talk about the p4/p4c technical details separately, but in regards to this PR, here is some overarching commentary:
- I am not sure we want to maintain a parallel copy of dash-pipeline P4 code. Normally I'd prefer to put in conditionals in the code and pass in an environment variable in the call to `p4c`.  This would reduce maintenance and make the differences obvious. Keeping two versions does the opposite. There's a tradeoff if the conditional code becomes unreadable, but I get the impression it won't be.
- Similarly we can add the new `make` target to `dash-pipeline/Makefile`
- To make this a bona fide build option and be able to add in CI/CD for regression checks (which seems worthwhile), we need to make the `p4c-dpdk` executable available to the dash dev environment. This might not be difficult. If you look inside the dockerfile which builds `p4c` for bmv2, you'll find we simply take a selected distro of p4c docker image from Dockerhub and selectively copy some binaries into our final docker image, to reduce size. This particular line of code could possibly be modified to copy the p4c-dpdk target: https://github.com/sonic-net/DASH/blob/2c785737a1bf2e483a0ac705ae89d9fcc0debb4d/dash-pipeline/dockerfiles/Dockerfile.p4c-bmv2#L43
We'd need to select an appropriate version of the source docker image, which is defined here: https://github.com/sonic-net/DASH/blob/2c785737a1bf2e483a0ac705ae89d9fcc0debb4d/dash-pipeline/dockerfiles/Dockerfile.p4c-bmv2#L6

Alternatively, we could build and maintain a `p4c-dpdk` image all by itself (either by compiling it, or copying/pruning like we do now) and leave the `p4c-bmv2` image alone.
 
Note, there's been work underway by @marian-pritsak  et al, to build p4c from scratch to support some of the new DASH-inspired enhancements. This might influence the approach taken.
Let's discuss this at the next WG. I can help with some of the workflows and logistics of making a consolidated codebase and tool docker image if that's the direction we agree on.

jafingerhut:
I agree we should discuss these things and decide how this should be made into a PR we would actually want to merge.  This PR is primarily publishing my work in progress in an easily findable place that others can check out if they are interested.

I also believe that a single P4 code base with a few more #ifdef's should be a reasonable way to go.  There are not that many different lines of code between what is there in the bmv2 directory, and my modified version.

dash-pipeline/dpdk-pna/README.md
jafingerhut: Fixed

chrispsommers:
Andy, thanks for the response, I sort of figured you were (as usual) providing a well-documented and stable vantage point from which to make subsequent decisions. We shall discuss!

chrispsommers:
Hi @jfingerh I submitted a PR against your https://github.com/jfingerh/DASH/tree/add-p4-dpdk-reference-code branch to add p4c-dpdk docker image and make targets to the workflow, please take a look.

mhanif:
Folks, let's avoid ifdef in our code.  The use of it may balloon up if we let this one in.  So avoiding it from the get go is highly recommended.  Instead, we should put the target specific code (bmv2 vs dpdk) in separate files and keep the common code in common files.  Since we will be compiling target specific code (p4c-bm2-ss vs p4c-dpdk), let the makefile pick the right set of files and do the trick.

jafingerhut:
@mhanif I'm OK with approaches that do not add more ifdef's to the code, but. you do realize there are dozens already in there now, I hope?

chrispsommers:
Per Andy:
> I also believe that a single P4 code base with a few more #ifdef's should be a reasonable way to go. There are not that many different lines of code between what is there in the bmv2 directory, and my modified version.

I think this carries a lot of weight. Maintaining parallel files carries its own burden, so it needs to be a balanced decision. And, the decision can be reconsidered in the future if `#ifdef`s become untenable.

lguohan:
i agree, better to consolidated into one file.

mhanif:
> I think this carries a lot of weight. Maintaining parallel files carries its own burden, so it needs to be a balanced decision. And, the decision can be reconsidered in the future if `#ifdef`s become untenable.

@chrispsommers agreed.

chrispsommers:
Whether or not we merge the code or keep two versions, we could also make a CI pipeline to catch regressions.

jfingerh:
Closing this PR in favor of a newer one that adds #ifdef's to the existing P4 code, instead of creating copies of the files as this PR does.  The newer PR is this one: https://github.com/sonic-net/DASH/pull/332

PR #332, opened 2023-02-02
Add #ifdef's so P4 code can target BMv2 or DPDK back ends

jfingerh:


jfingerh:
@chrispsommers I did not include your extra changes in this PR (sorry, just laziness -- I have not looked at your additions yet).

One change I made that might affect your changes is that I renamed the #define symbol to enable DPDK compilation from DPDK_PNA to ARCH_DPDK_PNA.

Also, if you want to compile for BMv2 target, you must now have `-DARCH_BMV2_V1MODEL` on the p4c command line, or some equivalent.

chrispsommers:
Hi @Andy,thanks, this is a step in the right direction. Once I can manually build this I can install the tooling.Can you please provide the command line you used to build it for dpdk?

Somehow you broke the bmv2 p4 compilation, it needs to be fixed before we can proceed.
https://github.com/sonic-net/DASH/actions/runs/4078881600/jobs/7029623145#step:4:24
You need to be able to run `make p4` locally and also verify it in the CI actions, will will trigger with any push affecting the P4 code.  You can look at your own "Actions" page to see CI results, but you need to enable actions in your GitHub project settings - they're diabled. (Normally you should not even submit a PR to the DASH repo until your own CI actions pass.) Feel free to ask for help on this.

jfingerh:
@chrispsommers Yep, I noticed I broke the build.  It is because of the new ARCH_BMV2_V1MODEL preprocessor symbol that must be defined, which I added to the p4c command line in dash-pipeline/Makefile in a later commit after we both saw the build failure.

jfingerh:
Here is the p4c-dpdk command line I used:
```
p4c-dpdk -DARCH_DPDK_PNA -DPNA_CONNTRACK --pp dash_pipeline.pp.p4 -o dash_pipeline.spec --arch pna --bf-rt-schema dash_pipeline.p4.bfrt.json --p4runtime-files dash_pipeline.p4.p4info.txt dash_pipeline.p4
```
Feel free to eliminate any of these sets of command line options if you want fewer output files, or adjust the output file names/directories to your liking:
+ `--pp dash_pipeline.pp.p4`
+ `--bf-rt-schema dash_pipeline.p4.bfrt.json`
+ `--p4runtime-files dash_pipeline.p4.p4info.txt dash_pipeline.p4`

jfingerh:
Whenever we get to the point of running packets through the DPDK software switch, the `.spec` file will be needed, for sure., so I would recommend keeping that one, at least.

chrispsommers:
Hi Andy, thanks, I see it passes now. Let me work on another PR to your branch to enable the tooling, so we can make this a complete submission. I'll mark this draft for now.

chrispsommers:
@jfingerh I made a https://github.com/jfingerh/DASH/pull/2 to this branch. Please enable your GitHub actions and accept the PR. THanks.

dash-pipeline/bmv2/dash_acl.p4
```diff
@@ -44,6 +51,42 @@ match_kind {
...
+//        pna_direct_counter = ## table_name ##_counter; \
+
+#define ACL_STAGE(table_name) \
+    @name(str(table_name##:dash_acl_rule|dash_acl)) \
+    table table_name { \
+        key = { \
+            meta. ## table_name ##_dash_acl_group_id : exact @name("meta.dash_acl_group_id:dash_acl_group_id"); \
+            meta.dst_ip_addr : LIST_MATCH @name("meta.dst_ip_addr:dip"); \
+            meta.src_ip_addr : LIST_MATCH @name("meta.src_ip_addr:sip"); \
+            meta.ip_protocol : LIST_MATCH @name("meta.ip_protocol:protocol"); \
```
mhanif, line 77: Shouldn't "protocol" be exact match?  Why is this mapping to ternery?

dash-pipeline/bmv2/dash_dpdk_pna.p4
```diff
@@ -0,0 +1,16 @@
...
+{
+    apply {
+    }
+}
+
+PNA_NIC(
+    dash_parser(),
+    dash_precontrol(),
+    dash_ingress(),
+    dash_deparser()) main;
```
mhanif, line 16: Since PNA is a "portable" architecture, why tie it to DPDK only (that is, why call this file dpdk_pna.p4).  We can always pick different targets.  Isn't it?  Putting portable architectures with the targets just defeats the purpose in my view.  This is true towards usage of DPDK target with PNA name elsewhere in this PR as well.

mhanif (changes requested):
Few comments inline with the suggested code changes.

dash-pipeline/bmv2/dash_acl.p4
```diff
@@ -17,12 +17,19 @@ match_kind {
 #define LIST_MATCH list
 #define RANGE_LIST_MATCH range_list
 #else
+#ifdef ARCH_BMV2_V1MODEL
 #define LIST_MATCH optional
 #define RANGE_LIST_MATCH optional
+#endif // ARCH_BMV2_V1MODEL
+#ifdef ARCH_DPDK_PNA
+#define LIST_MATCH ternary
```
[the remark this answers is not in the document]
jafingerhut: There are no lists of MAC addresses in the current P4 code.  Maybe that is because the P4 code is not yet functionally complete?  I did not write the code that I started from.  Maybe the original author knows?

dash-pipeline/bmv2/dash_acl.p4
jafingerhut: The original code used match_kind optional (either complete exact match, or wildcard on all bits).  Should it have been exact match?  I am not claiming what I am writing is the final version.  Again, perhaps the original author can answer why it was the way it was in the code before this PR?

dash-pipeline/bmv2/dash_dpdk_pna.p4
jafingerhut: I am happy to rename the file "dash_pna.p4" if you and other prefer that file name, and I can change the preprocessor symbol to ARCH_PNA instead of ARCH_DPDK_PNA, too.

Changing those things will not automatically mean that this code compiles on all PNA target devices, since this code is written to be a functional reference model for a CPU-based target.  Different PNA devices will have different subsets of possible PNA programs they can compile, and this program seems unlikely to be in the intersection of all of those.

jafingerhut:
@chrispsommers You recommended that I "enable your Github actions".  Is that some kind of per-branch Github setting somewhere?  Any link to directions on how I do that?

chrispsommers:
GitHub actions are configured on a per-repository basis. Here's what you can do:
Go to your top-level page in your DASH fork; Select the settings cog (rightmost widget next to Insights, same line as `<> code`,
select `Actions | General` in left pane, and select "Enable All actions..." radio button under `Actions permissions`. See example below from my fork. Enjoy your workflows!

image

chrispsommers (approved):
LGTM

dash-pipeline/bmv2/dash_acl.p4
mhanif: This could have been a cut-and-paste issue :-)  I am a co-author of some of this code.  In the 5-tuple, the IP-Proto (protocol field) should have been an exact match.

dash-pipeline/bmv2/dash_acl.p4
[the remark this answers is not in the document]
mhanif: Agreed that currently the only list where LIST_MATCH will use will be a list of IP prefixes and ternery match will do just fine.

dash-pipeline/bmv2/dash_acl.p4
jfingerh: Do ALL DASH rules always have an exact match on IP protocol?  That is, is it disallowed to attempt to create a rule that matches on IP SA and DA only, and wildcard everything else?

mhanif: This is a great point.  Perhaps, in order to cover the wildcard case, this field was marked ternery.

jfingerh: If you search for the first occurrence of the string "ACL_LEVEL1" on this page https://github.com/sonic-net/DASH/blob/main/documentation/general/sdn-pipeline-basic-elements.md you can see a couple of example configurations where the protocol contains `*`, which appears to be a wildcard match on the packet protocol field.  If those examples are intended to be supported by DASH, then it appears the data plane needs a way for protocol to be completely wildcarded, or completely exact match.  In P4, ternary is more general than that.  There is another match kind called optional that is exactly that.  It is not clear to me from that example whether the control plane would ever take advantage of the ability to configure more than one exact match protocol id for a single rule.

mhanif (approved):
LGTM

chrispsommers:
I removed draft status, looks ready to merge!

James92618:
Hi Andy,

BMv2(a P4 Soft Switch) target is expressed in "target, arch" tuple. when looking for compiling help,
    $ p4c --help-targets
    Supported targets in "target, arch" tuple:
    dpdk-psa
    bmv2-v1model
    bmv2-psa
    ebpf-v1model

When checking for p4c-dpdk command line help, architecture is listed all the way down, after --target.
    $p4c-dpdk
    [--Werror=expected] error: No input files specified
    p4c-dpdk: Compile a P4 program
    ...
    ...
    --target target                         Compile for the specified target device.
    --arch arch                             Compile for the specified architecture.

May I suggest changing from ARCH_DPDK_PNA to TARGET_DPDK_PNA? We misuse and confused with these words all the times. I think TARGET_xxx is a more precise definition.
When we start testing with running a Simple_Switch instance, simple_switch is a target, not an architecture.

jfingerh:
@James92618 I am all for having good names for these preprocessor symbols that people like.

If there is some way to achieve consensus on what they should be changed to among the people that are developing the reference code, please let me know what the result of that decision is.

chrispsommers:
> If there is some way to achieve consensus on what they should be changed to among the people that are developing the reference code, please let me know what the result of that decision is.

@jfingerh Is there any reason to hold this PR while this debate ensues, which more about p4c? I'd like to merge it.

chrispsommers:
> @chrispsommers Changing the names of some preprocessor symbols later is a fairly simple global search-and-replace on those strings either now before this PR is merged, or later after it is merged.
> 
> I am not familiar with the decision-making on what gets merged in this repo yet, so defer to others on whether such a change is made before or after.

Hi Andy, if you think the recommendations on the preprocessor name is reasonable, feel free to edit the files and do so. We can take our best shot now and revise later if we want to.

dash-pipeline/dockerfiles/Dockerfile.p4c-dpdk
```diff
@@ -0,0 +1,61 @@
+# This Dockerfile builds an image used to compile P4 programs for the bmv2 backend only
```
mhanif, line 1: I believe this comment needs to be modified since this Dockerfile is for DPDK.  Cut-and-paste typo?

chrispsommers: @mhanif you are correct, that's my mistake, thanks for catching! @jfingerh please fix this?

lguohan:
since we build the dpdk docker, is it possible to run our ptf thrift test against this new p4-dpdk model?

chrispsommers:
> since we build the dpdk docker, is it possible to run our ptf thrift test against this new p4-dpdk model?

Hi @lguohan the new docker is only the p4c **compiler** itself with the dpdk backend. It is added to the CI in this PR. It does not include any p4-dpdk backend **executable**  yet, that is the open task I was referring to in the meeting ("someone needs to break the ice on the dpdk backend"). The goal is to integrate the backend into the CI workflow and run all tests against it. Then we will have two parallel soft switches (bmv2 and dpdk).

jfingerh:
@James92618 @chrispsommers Commit 8 on this PR renames the two `ARCH_<name>` preprocessor symbols to `TARGET_<name>` instead.

dash-pipeline/dockerfiles/Dockerfile.p4c-dpdk
jfingerh: I have added commit 9 on this PR to fix this comment.

chrispsommers:
In today's WG meeting we debated freezing `p4c-dpdk` backend to a specific SHA, or leave it floating, i.e. `:latest`. We agreed to leave floating to act as a canary for breaking changes in p4c. We can freeze in the future.

PR #336, opened 2023-02-13
[ACL]: Optimize DASH ACL by introducing cluster tag

Pterosaur:


documentation/general/dash-sonic-hld.md
```diff
@@ -305,14 +305,31 @@ DASH_ACL_GROUP_TABLE:{{group_id}}
     "guid": {{string}}
 ```
 
+```
+DASH_ACL_SRC_TAG_TABLE:{{tag_name}}
+    "addresses": {{list of prefix}}
```
prsunny, line 310: please rename to prefix_list

documentation/general/dash-sonic-hld.md
```diff
@@ -305,14 +305,31 @@ DASH_ACL_GROUP_TABLE:{{group_id}}
     "guid": {{string}}
 ```
 
+```
+DASH_ACL_SRC_TAG_TABLE:{{tag_name}}
```
prsunny, line 309: Please rename to DASH_PREFIX_TAG_TABLE. One table is enough for both src and dst as controller does not have to repeat it. Also the same could be used in future for non-ACL cases. So lets not prefix ACL to the table name.

documentation/general/dash-sonic-hld.md
```diff
@@ -305,14 +305,31 @@ DASH_ACL_GROUP_TABLE:{{group_id}}
...
+    "addresses": {{list of prefix}}
+```
+
+```
+DASH_ACL_DST_TAG_TABLE:{{tag_name}}
```
prsunny, line 314: We dont need this table

documentation/general/dash-sonic-hld.md
```diff
@@ -305,14 +305,31 @@ DASH_ACL_GROUP_TABLE:{{group_id}}
...
+
 ```
 DASH_ACL_RULE_TABLE:{{group_id}}:{{rule_num}}
     "priority": {{priority}}
     "action": {{action}}
     "terminating": {{bool}}
     "protocol": {{list of protocols}}
-    "src_addr": {{list of address}}
-    "dst_addr": {{list of address}}
+    "src_tag_map": {{list of tag name}}
```
prsunny, line 329: `src_tag` should be sufficient

Pterosaur: Why don't we need the dst list?

prsunny: oh i meant src_tag and dst_tag instead of src_tag_map and dst_tag_map

documentation/general/dash-sonic-hld.md
```diff
@@ -311,8 +321,10 @@ DASH_ACL_RULE_TABLE:{{group_id}}:{{rule_num}}
     "action": {{action}}
     "terminating": {{bool}}
     "protocol": {{list of protocols}}
-    "src_addr": {{list of address}}
-    "dst_addr": {{list of address}}
+    "src_tag": {{list of tag name}}
```
prsunny, line 324: Mark this optional

Pterosaur: Should we have an explicit "optional" mark to "tag" field? IMHO, most of fields in DASH ACL RULE TABLE are optional. E.G. We can only set src_tag or dst_port and set other fields to be empty. So, should we also mark them?

dash-pipeline/bmv2/dash_acl.p4
```diff
@@ -29,6 +30,8 @@ match_kind {
     table table_name { \
         key = { \
             meta. ## table_name ##_dash_acl_group_id : exact @name("meta.dash_acl_group_id:dash_acl_group_id"); \
+            meta.dst_tag_map : ternary @name("meta.dst_tag_map:dst_tag"); \
```
marian-pritsak, line 33: match type: optional

dash-pipeline/bmv2/dash_pipeline.p4
```diff
@@ -242,6 +242,34 @@ control dash_ingress(inout headers_t hdr,
...
     }
 
+    action set_src_tag(tag_map_t tag_map) {
+        meta.src_tag_map = tag_map;
+    }
+
+    @name("src_tag|dash_tag")
+    table src_tag {
+        key = {
+            meta.src_ip_addr : lpm @name("meta.src_ip_addr:sip");
```
marian-pritsak, line 252: match type: prefix_list

marian-pritsak (changes requested):

dash-pipeline/bmv2/dash_acl.p4
Pterosaur: Why is here optional?
1. We would like a mask operation.
E.G. If the source IP of packet belong to tag 1 and 2, and this ACL rule is applied to tag 2 and 3, it should be matched. If another ACL rule is applied to tag 3 and 4, it should be mismatched.
2. If the tag was provided, the target rule could match all tags.
I'm not sure whether the optional match_kind can meet above behaviors.

documentation/general/dash-sonic-hld.md
```diff
@@ -283,7 +283,17 @@ pl_sip_encoding          = Private Link encoding for IPv6 SIP transpositions; Fo
...
-  
+
+```
+DASH_PREFIX_TAG_TABLE:{{tag_name}}
+    "prefix_list": {{list of prefix}}
+```
+
+```
+tag_name                  = STRING; unique tag name
+addresses                 = list of ip prefixes ',' separated
```
prsunny, line 294: add a statement that says, "valid to have empty list of prefixes"

documentation/general/dash-sonic-hld.md
```diff
@@ -325,10 +337,12 @@ priority                 = INT32 value  ; priority of the rule, lower the value,
 action                   = allow/deny
 terminating              = true/false   ; if true, stop processing further rules
 protocols                = list of INT ',' separated; E.g. 6-tcp, 17-udp; if not provided, match on all protocols
-src_addr                 = list of source ip prefixes ',' separated
-dst_addr                 = list of destination ip prefixes ',' separated
-src_port                 = list of range of source ports ',' separated
-dst_port                 = list of range of destination ports ',' separated
+src_tag_map              = list of source tag name ',' separated; if not provided, match on all source TAGs.
```
Pterosaur, line 340:
```suggestion
src_tag              = list of source tag name ',' separated; if not provided, match on all source TAGs.
```

dash-pipeline/bmv2/dash_acl.p4
Pterosaur: A case is if a prefix belong is to a tag, A, B, C, D (1111) and an ACL rule is applied to tag A, C(1010). 
If my understanding is correct, to the match_kind, optional. the mask should be 1111, which implies `tag (1111) & mask(1111) != target_tag(1010)`.
But to the match_kind, ternary, the mask will be 1010. So the rule will be matched due to `tag(1111) & mask(1010) = target_tag(1010)`.

documentation/general/dash-sonic-hld.md
```diff
@@ -310,11 +320,13 @@ DASH_ACL_RULE_TABLE:{{group_id}}:{{rule_num}}
     "priority": {{priority}}
     "action": {{action}}
     "terminating": {{bool}}
-    "protocol": {{list of protocols}}
-    "src_addr": {{list of address}}
-    "dst_addr": {{list of address}}
-    "src_port": {{list of range of ports}}
-    "dst_port": {{list of range of ports}}
+    "protocol": {{list of protocols}} (OPTIONAL)
+    "src_tag": {{list of tag name}} (OPTIONAL)
```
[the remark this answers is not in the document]
Pterosaur: 1. `specifying a src_tag with empty prefix must not match packets.` I add this description under the TAG table.
2. `When specifying a src_tag and src_prefix, it shall be a OR. Please clarify it in documentation`. I think this should be an AND instead of OR. If you want a OR, just to define another rule. 
There are three reasons:
1. The tag fields are same as other fields like protocol, port and so on. The relationship between src, protocol and port should be AND.
2. The TAG is for identifying a packets that belongs to a specific group. In the further, maybe we can identify a packets not only from address but also from other features.
3. I think if you want to implement `OR` in P4, we have to separate the TAG and ADDR to different ACL tables, which will increase the complexity of the ACL system.

If you accept my points, I can polish and clarify the descriptions for TAG.

documentation/general/dash-sonic-hld.md
Pterosaur: Done

documentation/general/dash-sonic-hld.md
```diff
@@ -283,7 +283,17 @@ pl_sip_encoding          = Private Link encoding for IPv6 SIP transpositions; Fo
 pl_underlay_sip          = Underlay SIP (ST GW VIP) to be used for all private link transformation for this ENI
 ```
 ### 3.2.4 ACL
-  
+
+```
+DASH_PREFIX_TAG_TABLE:{{tag_name}}
+    "prefix_list": {{list of prefix}}
```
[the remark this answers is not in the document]
Pterosaur: Done

mhanif:
@mukeshmv could you please add your scale related comments/questions to this review.  Thanks!

documentation/general/dash-sonic-hld.md
```diff
@@ -283,7 +283,19 @@ pl_sip_encoding          = Private Link encoding for IPv6 SIP transpositions; Fo
 pl_underlay_sip          = Underlay SIP (ST GW VIP) to be used for all private link transformation for this ENI
 ```
 ### 3.2.4 ACL
-  
+
+```
+DASH_PREFIX_TAG_TABLE:{{tag_name}}
```
mukeshmv, line 288: Can we add some scale numbers for the ACL tags in the HLD ? 
 - How many total tags ?
 - How many prefixes per tag ?
 - How many tags can each prefix belong to ?

KrisNey-MSFT:
@prsunny - may we have an update re: this one?  There will be work needed in SAI.

documentation/general/dash-sonic-hld.md
KrisNey-MSFT: We will need to update the HLD with the scale numbers.  
Possibly by sometime this week 3/16/2023
@prsunny

Pterosaur: Added the scale numbers of tags in the above scale table.

marian-pritsak:
No need to generated separate APIs for SRC/DST tags. Can be joined together like ACL APIs.

documentation/general/dash-sonic-hld.md
```diff
@@ -342,11 +359,13 @@ DASH_ACL_RULE_TABLE:{{group_id}}:{{rule_num}}
...
     "terminating": {{bool}}
-    "protocol": {{list of protocols}}
-    "src_addr": {{list of address}}
-    "dst_addr": {{list of address}}
-    "src_port": {{list of range of ports}}
-    "dst_port": {{list of range of ports}}
+    "protocol": {{list of protocols}} (OPTIONAL)
+    "src_tag": {{list of tag name}} (OPTIONAL)
+    "dst_tag": {{list of tag name}} (OPTIONAL)
+    "src_addr": {{list of prefix}} (OPTIONAL)
```
vincent-xs, line 365: the documentation switched from address to prefix (prefix is definitely what we would expect). 
BUT the P4 model is still generating IP address, not prefix. could the  P4 code be fixed to match the documentation?

    /**
     * @brief Optional matched key dip
     *
     * @type sai_ip_address_t
     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
     */
    SAI_DASH_ACL_RULE_ATTR_DIP,

    /**
     * @brief Optional matched key sip
     *
     * @type sai_ip_address_t
     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
     */
    SAI_DASH_ACL_RULE_ATTR_SIP,

dash-pipeline/bmv2/dash_pipeline.p4
```diff
@@ -268,6 +268,34 @@ control dash_ingress(
...
+        actions = {
+            set_src_tag;
+        }
+    }
+
+    action set_dst_tag(tag_map_t tag_map) {
+        meta.dst_tag_map = tag_map;
+    }
+
+    @name("dst_tag|dash_tag")
```
vincent-xs, line 289: I am not sure how the SAI API is derived from the  P4 code, but I noticed that we have separate APIs to create src tag vs dst tag in sai_dash_tag_api_t
why is this the case? this adds a lot of redundancy.

we should be able to define tags and then use those for src or destination.

this would be similar to ACL groups: we just create a group and then use is for ingress or egress.

documentation/general/dash-sonic-hld.md
Pterosaur: Have you tried to manually define the DASH_MATCH in the file, dash_acl.p4?
```c++
#define DASH_MATCH  // Please explicitly define this line in dash-pipeline/bmv2/dash_acl.p4

#ifdef DASH_MATCH
#define LIST_MATCH list
#define RANGE_LIST_MATCH range_list
#else
#ifdef TARGET_BMV2_V1MODEL
#define LIST_MATCH optional
#define RANGE_LIST_MATCH optional
#endif // TARGET_BMV2_V1MODEL
#ifdef TARGET_DPDK_PNA
#define LIST_MATCH ternary
#define RANGE_LIST_MATCH range
#endif // TARGET_DPDK_PNA
#endif
```
If you define it, you will get the correct SAI header.
```
    /**
     * @brief List matched key dip
     *
     * @type sai_ip_prefix_list_t
     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
     */
    SAI_DASH_ACL_RULE_ATTR_DIP,

    /**
     * @brief List matched key sip
     *
     * @type sai_ip_prefix_list_t
     * @flags MANDATORY_ON_CREATE | CREATE_ONLY
     */
    SAI_DASH_ACL_RULE_ATTR_SIP,
```

dash-pipeline/bmv2/dash_pipeline.p4
svshah-intel: Could you have a scenario where eg. 10.0.0.0/8 -> tag1, 10.1.0.0/16 -> tag2.. How do you expect lpm match here to resolve and assign tags?

documentation/general/dash-sonic-hld.md
```diff
@@ -357,13 +376,15 @@ priority                 = INT32 value  ; priority of the rule, lower the value,
 action                   = allow/deny
 terminating              = true/false   ; if true, stop processing further rules
 protocols                = list of INT ',' separated; E.g. 6-tcp, 17-udp; if not provided, match on all protocols
-src_addr                 = list of source ip prefixes ',' separated
-dst_addr                 = list of destination ip prefixes ',' separated
-src_port                 = list of range of source ports ',' separated
-dst_port                 = list of range of destination ports ',' separated
+src_tag                  = list of source tag name ',' separated; if not provided, match on all source TAGs or no TAG.
```
svshah-intel, line 379: what does it mean by "match on all source TAGs or no TAG" ?

KrisNey-MSFT: Intention is if the attribute is not provided, it is like any other ACL rule.  If not provided, wildcard match.  Match on ANY tag or NO tag (reword)?  
@Pterosaur @prsunny

marian-pritsak:
> > No need to generated separate APIs for SRC/DST tags. Can be joined together like ACL APIs.
> 
> Hi @marian-pritsak , the APIs were automatically generated from P4. As you know, we only have a list of TAG-Prefix. So we don't know if the prefix is for SRC or DST to an actual packet. So, we have to try the SRC and DST both for tagging the packet. I have no idea about how to combine two fields(SRC/DST) matching into one key in P4. Like
> 
> ```
>     @name("dst_tag|dash_tag")
>     table dst_tag {
>         key = {
>             meta.dst_ip_addr/meta.src_ip_addr : lpm @name("meta.dst/src_ip_addr:dip");
>         }
>         actions = {
>             set_tag;
>         }
>     }
> ```
> 
> If you have any solution, I'm very glad to learn it.

You need to keep two tables in P4 code, but unite them under one API name, like in the example you gave with this annotation `@name("dst_tag|dash_tag")`

prsunny:
@Pterosaur , can you please resolve conflict?

documentation/general/dash-sonic-hld.md
vincent-xs: @marian-pritsak 

is that something that could be taken care of so we align things up?
thanks

marian-pritsak:
> > > No need to generated separate APIs for SRC/DST tags. Can be joined together like ACL APIs.
> > 
> > 
> > Hi @marian-pritsak , the APIs were automatically generated from P4. As you know, we only have a list of TAG-Prefix. So we don't know if the prefix is for SRC or DST to an actual packet. So, we have to try the SRC and DST both for tagging the packet. I have no idea about how to combine two fields(SRC/DST) matching into one key in P4. Like
> > ```
> >     @name("dst_tag|dash_tag")
> >     table dst_tag {
> >         key = {
> >             meta.dst_ip_addr/meta.src_ip_addr : lpm @name("meta.dst/src_ip_addr:dip");
> >         }
> >         actions = {
> >             set_tag;
> >         }
> >     }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If you have any solution, I'm very glad to learn it.
> 
> @Pterosaur You need to keep two tables in P4 code, but unite them under one API name, like in the example you gave with this annotation `@name("dst_tag|dash_tag")`

@Pterosaur after considering this, I think it's better to generate separate APIs for src and dst tags, which will allow implementations keep two different LPM tables for a better utilization of resources.

marian-pritsak (approved):

documentation/general/dash-sonic-hld.md
KrisNey-MSFT: @vincent-xs  will look at SAI Master for generation code 
Difference between APIs for SAI and bmv2 defined match types for SIP and DIP
By default not-enabled (hidden compile to check for possibly)?
@Pterosaur possibly can also check the 'define'

documentation/general/dash-sonic-hld.md
```diff
@@ -359,13 +378,15 @@ priority                 = INT32 value  ; priority of the rule, lower the value,
...
```
KrisNey-MSFT, line 381: @Pterosaur and @prsunny 
Instead of this:  '_match on all source TAGs or no TAG_'
Could we have a rewording to:  '_match on ANY tag or NO tag_'?

Per Community conversation, the intention is that if the attribute is not provided, it is like any other ACL rule.  If not provided, it will be a wildcard match.  Could we have it reworded similar to the above?

Pterosaur: OK, I will submit a PR to polish it.

Pterosaur: Change it in PR: https://github.com/sonic-net/DASH/pull/359

PR #339, opened 2023-02-15
Inbound and Bidirectional functional cases 

desaimg1:
This test covers following functional tests :
UDP traffic verified on inbound direction
UDP traffic verified on bidirectional

Notes:
This test passed with real hardware and failing with BMv2. Needs fix with BMv2 on resolving routes.


test/test-cases/functional/saic/test_vm_to_vm_commn_acl_inbound.py
```diff
@@ -0,0 +1,179 @@
...
+import time
+import pytest
+import sys
+sys.path.append("../utils")
+import snappi_utils as su
+
+current_file_dir = Path(__file__).parent
+
+"""
+This covers following scenario :
```
aputriax, line 13: Could you please extend test description here and in two other test modules. It's difficult to understand what is expected when part of the configuration is in a separate file and another part is here but "smeared" in the test function code.
I would suggest adding a function kind of build_vxlan_packet (or flow) which could be used here and which can help reduce number of code lines in the test cases.
In that case it will be easier to comprehend what is sent and what is expected.

test/test-cases/functional/saic/test_vm_to_vm_commn_udp_bidir.py
```diff
@@ -0,0 +1,295 @@
...
+        res2 = su.check_flow_tx_rx_frames_stats(dataplane, f2.name)
+        res3 = su.check_flow_tx_rx_frames_stats(dataplane, f3.name)
+        res4 = su.check_flow_tx_rx_frames_stats(dataplane, f4.name)
+        
+        dataplane.teardown()
+
+        # Validate test result  
+        print("\n======= Print Test Results  =======")        
+        print("res1 and res2 and res3 and res4 is {} {} {} {}".format(res1, res2, res3, res4))
+        assert res1, "Traffic test failure"
```
aputriax, line 275: It would be good if you will identify which traffic item failed - endpoints/direction/in or outbound
Also you can build and print a table that can help to debug (see prettytable module).

aputriax (changes requested):

test/test-cases/functional/saic/test_vm_to_vm_commn_acl_inbound.py
```diff
@@ -0,0 +1,179 @@
...
+@pytest.mark.xfail(reason="https://github.com/sonic-net/DASH/issues/345")
+class TestAclInbound:
+    @pytest.fixture(scope="class")
+    def setup_config(self):
+        """
+        Fixture returns the content of the file with SAI configuration commands.
+        scope=class - The file is loaded once for the whole test class
+        """
+        current_file_dir = Path(__file__).parent
+        with (current_file_dir / 'config_inbound_setup_commands.json').open(mode='r') as config_file:
```
aputriax, line 63: To improve readability you can do one of two things:

OPTION no.1:
Use variables in the JSON configuration like `{DPU_VTEP_IP}` and when load it do the following
```
config = json.loads(open(file_path).read().format(**globals()))
```

OPTION no.2:
Convert configs to *.py and move there all constants and import this config.py in the test case.

*How it can help?*
- You won't need to keep in track test's code and config files to have same IPs, etc.
- Both file will use the same variable names and it will improve readability.

test/test-cases/utils/snappi_utils.py
```diff
@@ -0,0 +1,62 @@
+def config_l1_properties(dataplane, usd_speed):
```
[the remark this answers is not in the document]
desaimg1: Addressed

test/test-cases/utils/snappi_utils.py
```diff
@@ -0,0 +1,62 @@
...
+def check_bgp_neighborship_established(dataplane):
+    req =dataplane.api.metrics_request()
+    req.bgpv4.column_names = ["session_state"]
+    results = dataplane.api.get_metrics(req)
+    ok = []
+    for r in results.bgpv4_metrics:
+        ok.append(r.session_state == "up")
+    return all(ok)
+
+def check_ping(dataplane, ip_obj_name, ip, ip_type="ipv4"):
```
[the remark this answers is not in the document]
desaimg1: Modified

test/test-cases/functional/saic/test_vm_to_vm_commn_acl_inbound.py
```diff
@@ -0,0 +1,179 @@
...
+        inner_udp.src_port.value = 20000
+        inner_udp.dst_port.value = 10000
+
+        dataplane.set_config()
+
+        # Verify traffic
+        print("\n======= Verify traffic with allowed packets passing =======")
+        print("\n======= Start traffic =======")
+        su.start_traffic(dataplane, f2.name)
+        time.sleep(TRAFFIC_SLEEP_TIME)
```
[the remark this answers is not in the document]
desaimg1: Modified

test/test-cases/functional/saic/test_vm_to_vm_commn_udp_bidir.py
desaimg1: if we use table modules, that is  to be installed in pipeline.
Added detailed print statement instead

test/test-cases/utils/snappi_utils.py
```diff
@@ -0,0 +1,62 @@
...
+    req = dataplane.api.ping_request()
+    if ip_type == "ipv4":
+        p1 = req.endpoints.ipv4()
+    else:
+        p1 = req.endpoints.ipv6()
+    p1.src_name = ip_obj_name
+    p1.dst_ip   = ip
+    responses = dataplane.api.send_ping(req).responses
+    for resp in responses :
+        if resp.src_name == ip1.name and resp.dst_ip == ip:
```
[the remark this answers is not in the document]
desaimg1: Modified

test/test-cases/functional/saic/test_vm_to_vm_commn_acl_inbound.py
desaimg1: Modified

test/test-cases/functional/saic/test_vm_to_vm_commn_acl_inbound.py
desaimg1: This is breaking the test.. Will take care of this in the next cases

test/test-cases/functional/saic/test_vm_to_vm_commn_udp_bidir.py
```diff
@@ -0,0 +1,299 @@
...
+        vxlan1.reserved1.value = 0
+
+        inner_eth1.src.value = INNER_SRC_MAC
+        inner_eth1.dst.value = INNER_DST_MAC
+
+        inner_ip1.src.value = ENI_IP   #ENI
+        inner_ip1.dst.value = NETWORK_IP1  #world
+
+        inner_udp1.src_port.value = 10000
+        inner_udp1.dst_port.value = 20000
```
aputriax, line 127: Please consider using dataplane's methods add_flow(), add_ethernet_header(), add_*_header() or add add_vxlan_flow() function to your snappi_utils.py.
As I mentioned in my first comment, the test code is long and not easy to comprehend.

test/test-cases/functional/saic/test_vm_to_vm_commn_acl_inbound.py
aputriax: Why? You are not changing the content, just loading it in a different way.

test/test-cases/functional/saic/test_vm_to_vm_commn_udp_bidir.py
aputriax: thanks!

test/test-cases/utils/snappi_utils.py
```diff
@@ -0,0 +1,84 @@
...
+    req.bgpv4.column_names = ["session_state"]
+    results = dataplane.api.get_metrics(req)
+    ok = []
+    for r in results.bgpv4_metrics:
+        ok.append(r.session_state == "up")
+    return all(ok)
+
+def check_ping(dataplane, ip_obj_name, ip, addr_family="ipv4"):
+    """
+    This function will verify ping connectivity between TGEN and DUT
```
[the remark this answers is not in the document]
desaimg1: This one still needs enhancement as it is not used currently.. So we can enhance in further usages of this API

chrispsommers:
@aputriax please review/resolve is possible, thanks

aputriax:
@chrispsommers I have a single unresolved concern regarding this PR. Currently, variables in py file (like DPU_VTEP_IP, ENI_VTEP_IP, NETWORK_VTEP_IP) should be in sync with the corresponding values in JSON configuration. This makes test hard to maintain and not clear enough for understanding. I suggested two ways to resolve this. Probably the easiest way is to convert JSON to PY and just import configuration dictionary from it to the test module. In this case the constants mentioned above (and all others) can be used directly in configuration dictionary. This will simplify maintenance, and eliminate possibility of errors during future test updates.
I think this is important since these are the very first tests and we should provide a good example that are clear to understand and easy to maintain.

chrispsommers:
> @chrispsommers I have a single unresolved concern regarding this PR. Currently, variables in py file (like DPU_VTEP_IP, ENI_VTEP_IP, NETWORK_VTEP_IP) should be in sync with the corresponding values in JSON configuration. This makes test hard to maintain and not clear enough for understanding. I suggested two ways to resolve this. Probably the easiest way is to convert JSON to PY and just import configuration dictionary from it to the test module. In this case the constants mentioned above (and all others) can be used directly in configuration dictionary. This will simplify maintenance, and eliminate possibility of errors during future test updates. I think this is important since these are the very first tests and we should provide a good example that are clear to understand and easy to maintain.

@aputriax Thanks for the feedback, always welcome! We'll discuss internally.

desaimg1:
> @aputriax I had modified file as per your suggestion. Please review. 

Thanks,
Desai

aputriax (approved):

chrispsommers (approved):
LGTM

PR #350, opened 2023-03-15
[doc]: Add protobuf description for dash app_db

Pterosaur:
Add protobuf doc for DASH APP DB


documentation/general/dash-sonic-hld.md
```diff
@@ -447,82 +447,159 @@ overlay_dip              = ip_address                ; overlay dst ip if routing
...
+|                       |              | priority        | sai_inbound_routing_entry_t.priority            |                                               |
+|                       |              | protocol        |                                                 |                                               |
+|                       |              | vnet            | SAI_INBOUND_ROUTING_ENTRY_ATTR_SRC_VNET_ID      |                                               |
+|                       |              | pa_validation   | SAI_INBOUND_ROUTING_ENTRY_ATTR_ACTION           | use PA_VALIDATE if true                       |
+|                       |              | metering_bucket |                                                 |                                               |
+
+
+### 3.2.11 Protobuf encoding
+
+For saving memory consumption, the DASH table of APP_DB could be encoded as protobuf
```
chrispsommers, line 529: Are you proposing a new bespoke, hand-generated schema for just this case? I'm concerned about one-off schema for a narrow problem. It would be nice to have a general approach to defining protobuf schema for any DB accesses, which can be auto-generated from some other canonical schema.

Pterosaur: Hi Chris, thanks for your comment.
In my opinion, we only have a schema to define table content that indeed was in this doc. But we don't have any explicit description for the table implementation.
In a typical SONiC convention, all tables are stored in the Redis as clear text. But I did an experiment that shows memory consumption isn't acceptable in the DASH scenario. So, we need to seek a method to store all tables with lower memory usage.
So, Protobuf that I had surveyed is a good format that has a tradeoff between memory consumption and maintainability.

In the summary, the question is what the "canonical schema" is in your comment. If we didn't have a canonical schema to descript the storing structure, why this Protobuf cannot be that?

chrispsommers: Hi Ze, thanks for the explanation, I didn't understand the full context. I guess what you're saying is, this *is* the canonical schema and it just happens to look different from schema we've defined in the past. I agree that protobuf has many desirable properties. Do you intend to propose similar schema for all of DASH config?

Pterosaur: Yes, we would like to store all APP_DB of DASH as Protobuf. And the user can leverage the gNMI(https://github.com/sonic-net/sonic-gnmi) to interop with SONiC by the Protobuf we defined.

KrisNey-MSFT:
Per the DASH Community meeting, it would be nice to have a diagram/drawing to go along with this PR for visualization (if possible)?

Pterosaur:
> Per the DASH Community meeting, it would be nice to have a diagram/drawing to go along with this PR for visualization (if possible)? And if document is sanitized, @Pterosaur - would it be possible to publish the #'s presented?

The diagram will be in another PR for the whole architecture and data flow. Because the whole data flow are still investigating.
According to the meeting discussion, I will public a Excel file for memory usage caculation. After that, I will publish this PR .

Pterosaur:
@prsunny

ganglyu (approved):
LGTM

prsunny (approved):

PR #353, opened 2023-03-24
Update SAI-C client docker

andriy-kokhan:

PR #355, opened 2023-03-30
Skipping SAI-C TCs affected by issues/345

andriy-kokhan:
This PR fixes SAI-C CI issue caused by https://github.com/sonic-net/DASH/pull/339
In case TC setup process fails at some step, we should not try to remove all entries on teardown because some of them were not created. For known issues, the simplest solution is to replace `xfail` with `skip`.

### DIFF
--- a/test/test-cases/functional/saic/test_vm_to_vm_commn_acl_inbound.py
+++ b/test/test-cases/functional/saic/test_vm_to_vm_commn_acl_inbound.py
@@ -38,7 +38,7 @@ Topology Used :
 ###############################################################
 
 
-@pytest.mark.xfail(reason="https://github.com/sonic-net/DASH/issues/345")
+@pytest.mark.skip(reason="https://github.com/sonic-net/DASH/issues/345")
 class TestAclInbound:
     @pytest.fixture(scope="class")
     def setup_config(self):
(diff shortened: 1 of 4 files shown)


andriy-kokhan:
> Hi @andriy-kokhan, any idea why your branch https://github.com/andriy-kokhan/DASH/tree/sai-c-docker-update without this fix passed CI https://github.com/andriy-kokhan/DASH/actions/runs/4523715606/jobs/7967050169 before we merged #353 but it failed in CI?

I believe it's because CI was using older SAI-C docker which has an issue with catching remove failures. E.g., https://github.com/opencomputeproject/SAI-Challenger/issues/30
Which is expected to be fixed in the updated SAI-C docker.

chrispsommers:
@andriy-kokhan why would CI and your PR branch (which was run in your own fork's CI) have different sai-c dockers? They are based on the same codebase, only the test cases changed.

andriy-kokhan:
> @andriy-kokhan why would CI and your PR branch (which was run in your own fork's CI) have different sai-c dockers? They are based on the same codebase, only the test cases changed.

My PR passed CI before #339 was merged. The TCs from #339 had a few setup/teardown issues that were not detected by old version of SAI-C docker. So, CI on #339 also passed. Then, #339 was merged. Then, my PR was merged and post merge CI was started which pushed new SAI-C docker into the registry. Then, TCs failed because on post merge CI because of the issues introduced by #339. That's how I understand this.

chrispsommers:
@andriy-kokhan thanks, that makes sense. It was @desaimg1 's new test cases which failed based on the order of merge accepts etc. Thanks for addressing this quickly!

chrispsommers (approved):
LGtM, thanks!

PR #368, opened 2023-05-17
update saithrift-bld Dockerfile to accept the ubuntu image SHA as ARG

vincent-xs:
when libsai.so is created with a compiler version that is newer than what is run within the saithrift builder docker container, we may run into unresolved linkage error (such as unresolved __libc_single_threaded).
however, if the compiler is older, we are fine.

so we need to upgrade the distro used within the saithrift build docker image.
more generically, this change allows to provide the ubuntu SHA to be used by docker to build the image from the calling Makefile.

by default, we'll use a 22.04 ubuntu version, but this can be overridden by setting AMD64_UBUNTU_SHA256 to the wanted hash.

### DIFF
--- a/dash-pipeline/Makefile
+++ b/dash-pipeline/Makefile
@@ -73,6 +73,8 @@ DOCKER_RUN := docker run \
 	--rm \
 	$(DOCKER_FLAGS)
 
+SHA1SUM := sha1sum | awk '{print substr($$1,0,11);}'
+
 SAI/SAI: sai-submodule
 
 sai-submodule:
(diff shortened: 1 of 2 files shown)


chrispsommers (changes requested):
Please add some more details in description: motivation, brief summary of changes, usage. Also look at suggestion for dependencies of docker images. Thanks.

dash-pipeline/Makefile
```diff
@@ -238,7 +240,7 @@ kill-switch:
 # SAI-THRIFT SERVER TARGETS
 ###############################
 
-saithrift-server:
+saithrift-server: docker-saithrift-bldr
```
chrispsommers, line 243: This technique (as used [here](https://github.com/chrispsommers/DASH/blob/eced43dbd14aaed2f220225cf8eb3329fc629928/dash-pipeline/Makefile#L401)) might eliminate redundant builds:
```
docker-saithrift-client-bldr-image-exists:
	docker images --format "{{.Repository}}:{{.Tag}}" |grep $(DOCKER_SAITHRIFT_CLIENT_BLDR_IMG) || make docker-saithrift-client-bldr
...
docker-saithrift-client: docker-saithrift-client-bldr-image-exists
```

vincent-xs: incorporated

chrispsommers: I don't see it, did you push a change?

chrispsommers (approved):

PR #369, opened 2023-05-18
update to scale numbers

mgheorghe:
- background tcp flows 15M (was 1M)
- background udp flows 15M (was 1M)
- background traffic packet size reduced to 500B from 1500B
- total flow capacity reduced by half (down to 32M from 64M) -
Signed-off-by: Mircea Dan <mircea-dan.gheorghe@keysight.com>


lguohan:
@mgheorghe , i am reverting this pr. please reformat the pr, it is unclear what exactly change this pr contains.

lguohan:
@mgheorghe , why do i need to do rich diff? why normal text diff won't work?

PR #372, opened 2023-05-24
Update ixia-c/snappi to fix https://github.com/sonic-net/DASH/issues/357 (TLS 1.0/1.1 security concerns)

chrispsommers:
- Update ixia-c controller docker version to remove TLS1.0/1.1. This also pulls along newer versions of ixia-c traffic engine and snappi SDK.
- Update SAI Challenger submodule, which got same updates to snappi
- Improved some Makefile dependencies for docker images which use base images. This tests whether base images are available in the build env and if not, pull from ACR or build on-demand, once.


chrispsommers:
HI @Pterosaur are you able to review this fix? Thanks.

PR #365, opened 2023-05-12
Adding support for metering

vijasrin:
- Outbound metering P4 model changes in BMV2
- Changes to support the SAI api generation for metering tables
- Skipping failing SAIC tests till the metadata has been updated

PR #378, opened 2023-05-31
Updated SAI-Challenger to latest main as of May 30

andriy-kokhan:
This new version of SAI-Challenger auto-generates `sai_thrift_metadata.py` in run-time.
Fixes Issue https://github.com/sonic-net/DASH/issues/374


chrispsommers:
Hi @andriy-kokhan thanks for this. If this gets merged, would this allow @vijasrin to re-enable some of the tests which are skipped in https://github.com/sonic-net/DASH/pull/365 (except ones requiring changes to dpugen)?

chrispsommers (approved):
LGTM, thanks.

andriy-kokhan:
> Hi @andriy-kokhan thanks for this. If this gets merged, would this allow @vijasrin to re-enable some of the tests which are skipped in PR #365 (except ones requiring changes to dpugen)?

This PR enables metadata autogen. But one more improvement is needed - SaiObjType autogen which is addressed in https://github.com/sonic-net/DASH/pull/381.  With these two, some of the tests can be re-enabled. To do that, we will probably make another PR into #365.

chrispsommers:
Thanks for the clarification.

PR #380, opened 2023-05-31
check that all files have linux line endings before commit

mgheorghe:
added pre-commit and pre-commit ci to check for file line endings and update them to linux if needed

to be merged in after https://github.com/sonic-net/DASH/pull/379 gets merged


chrispsommers:
Merging per WG consensus this morning.

PR #385, opened 2023-06-07
Update dash-sonic-hld.md with Tunnel Scale

KrisNey-MSFT:
Adding for @prsunny to review


KrisNey-MSFT:
DASH is more flexible, current DASH covers this, marking as DONE
Do not need to Merge

PR #396, opened 2023-06-30
Fix ENI lookup ordering

marian-pritsak:


marian-pritsak:
Fixes https://github.com/sonic-net/DASH/issues/382

chrispsommers (approved):
LGTM

KrisNey-MSFT:
Covered on Behavioral Model call; moved table up.
Marian to sync, then we will merge.

PR #400, opened 2023-07-11
[Sonic] Update HLD for ACL tagging and metering

prsunny:
1. Clarify metering precedence
2. Change ACL tag bitmap to list


vijasrin (approved):
lgtm

KrisNey-MSFT:
Use a list instead of a bitmap for ACL tagging
Metering - resolve / clarify priority across policy routing and mapping

PR #410, opened 2023-07-31
Update fixed apis

kcudnik:
Update fixed apis, and add missing ones

### DIFF
--- a/dash-pipeline/SAI/templates/saifixedapis.cpp.j2
+++ b/dash-pipeline/SAI/templates/saifixedapis.cpp.j2
@@ -33,16 +33,6 @@ extern "C" {
 
 using namespace std;
 
-sai_status_t sai_create_switch_dummy(
-        _Out_ sai_object_id_t *switch_id,
-        _In_ uint32_t attr_count,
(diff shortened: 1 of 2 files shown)


kcudnik:
FYI @Pterosaur @lguohan

dash-pipeline/SAI/templates/saifixedapis.cpp.j2
```diff
@@ -79,14 +79,14 @@ sai_status_t sai_get_switch_attribute(
             return SAI_STATUS_SUCCESS;
         
         case SAI_SWITCH_ATTR_PORT_LIST:
-            // make a tmp port list, saiserver will free the memory
-            objlist = (sai_object_id_t *)malloc(sizeof(port_list));
```
kcudnik, line 83: this is memory leak, caller should provide memory for placing objects in the .list member

lguohan (approved):

dash-pipeline/SAI/templates/saifixedapis.cpp.j2
chrispsommers: Good catch!

PR #416, opened 2023-08-02
[SAI] Refactor libsai and add propper logger

kcudnik:
Add proper logger class for logging inside libsai, for now data will be logged into stderr and syslog, later on this could be changed to select only one output. Logger is simpler version of sonic-swss-common/common/logger.

Many changes are done in refactoring:
- internal generated functions are marked static
- whitespace on line ends are removed
- compiler warning flags are added
- unnecessary includes are removed
- DASH_LOG_ENTER macro added on every function
- return proper sai statuses instead of 0/-1
- api implementations are prefixed dash_ and marked as DASH_PRIVATE to not export outside library
- extern C added to sai.h to make proper C bindings instead of C++
- make proper sai api call using sai_api_query for vnet_out.cpp test, since previous one does not compile after moving functions private
- prefix macros using DASH_
- change fprintf stderr to new DASH logger
- return SAI_STATUS_NOT_SUPPORTED instead of SAI_STATUS_SUCCESS on sai_*get api when attribute is not implemented (suspended since it's causing ptf test to fail)
- return not implemented on global sai_* apis when api is actually not implemented
- mutex will use lock_guard instead of manual lock/unlock
- update gitignore to skip temporary files


kcudnik:
@marian-pritsak @chrispsommers @Pterosaur @lguohan kindly please review

kcudnik:
if sai_get_switch_attribute returns SAI_STATUS_NOT_SUPPORTED for unknown attribute, ptf tests skips and goes to wait forever

```
saidashacl.SaiThriftDashAclTest ... skipped 'SkipTest on expected error. sai_thrift_get_switch_attribute with errorcode: -2 error: sai_thrift_exception(status=-2)'
----------------------------------------------------------------------
Ran 1 test in 0.005s
OK (skipped=1)
```

@Pterosaur @marian-pritsak @chrispsommers please help fix that, if attribute is not supported then those apis should return not supported, returning success will cause for client to read some garbage from memory

kcudnik:
+ @prsunny for review

.gitignore
```diff
@@ -1,3 +1,5 @@
+**~
+**.swp
```
chrispsommers, line 2: `vi` user ;^)

kcudnik: vim XD

dash-pipeline/SAI/templates/saifixedapis.cpp.j2
```diff
@@ -2,136 +2,130 @@
...
-sai_status_t sai_get_switch_attribute(
+static sai_status_t dash_sai_get_switch_attribute(
         _In_ sai_object_id_t switch_id,
         _In_ uint32_t attr_count,
-        _Inout_ sai_attribute_t *attr_list) {
+        _Inout_ sai_attribute_t *attr_list)
+{
+    DASH_LOG_ENTER();
+
+    DASH_LOG_NOTICE("get_switch_attribute()");
```
chrispsommers, line 50: Hhy not `dash_sai_get_switch_attribute()`?
Also, why both LOG calls, aren't they redundant?

dash-pipeline/SAI/templates/saifixedapis.cpp.j2
```diff
@@ -2,136 +2,130 @@
...
+
+                break;
+
+            default:
+
+                DASH_LOG_WARN("attr %d is NOT SUPPORTED, but returning SAI_STATUS_SUCCESS", attr->id);
+
+                memset(&attr->value, 0, sizeof(attr->value)); // clear potential caller garbage
+
+                break; // TODO FIXME should return NOT SUPPORTED
```
chrispsommers, line 117: If I'm not mistaken, all these fixed APIs were added in order to support SAI-PTF discovery functions, even if they aren't really "supported". If you return other that SUCCESS it might break all PTF testing. Try it and see.

dash-pipeline/SAI/templates/saifixedapis.cpp.j2
kcudnik: leftover from fprintf(stderr, "sai_get_switch_attribute()\n"); DASH_LOG_ENTER(); is logging always on debug, i can change the name

dash-pipeline/SAI/templates/saifixedapis.cpp.j2
```diff
@@ -199,86 +205,111 @@ static sai_port_api_t sai_port_api_impl = {
...
         _In_ const sai_service_method_table_t *services)
 {
+    DASH_LOG_ENTER();
+
     const grpc::string _grpcTarget = "0.0.0.0:9559";
     char test_json[] = "/etc/dash/dash_pipeline.json";
     char test_proto_json[] = "/etc/dash/dash_pipeline_p4rt.txt";
     int dev_id = 0;
 
+    DASH_LOG_NOTICE("GRPC call SetForwardingPipelineConfig %s => %s, %s", _grpcTarget.c_str(), test_json, test_proto_json);
```
chrispsommers, line 309: Duplicated below?

dash-pipeline/SAI/templates/saifixedapis.cpp.j2
kcudnik: yes, i already tried that, and those tests fail, but ideally i would like to return here not supported, since in syncd we don't want success on all non existing attributes, we can handle them in different way, but if i return not supported here, then i dont know how to fix ptf tests

dash-pipeline/SAI/templates/saifixedapis.cpp.j2
kcudnik: yes, removing duplicate

chrispsommers (approved):
This is a really good overhaul, thanks! See minor comments.

dash-pipeline/SAI/templates/saifixedapis.cpp.j2
kcudnik: i think we could add dummy attributes that are used in ptf test cases, or mitigate that on saithrift

chrispsommers: > add dummy attributes that are used in ptf test cases

That seems reasonable.

kcudnik: now how to find all attributes that are used by test ptf test cases? :P

PR #418, opened 2023-08-02
Use shared_ptr instead of new/delete to prevent memory leak

kcudnik:
Let's try to avoid new/delete especially where goto is used in generated headers, it will be hard to track those pointers


kcudnik:
i tested this before, and it failed on grpc, somehow logic is breaking something somewhere, not sure where yet

kcudnik:
ok it passed we are good to go :)

chrispsommers (approved):
LGTM

PR #419, opened 2023-08-02
[SAI] Add support for dash configuration from sai interface

kcudnik:
Adding this support will allow syncd to easy pass configuration parameters to dash libsai using service method table


chrispsommers (approved):
Code LGTM but I am not very familiar with the usage of these APIs. Can you give an example how they might be useful for DASH?

kcudnik:
for dash it should be transparent, main use case is for syncd, for example we load profile.ini file which configure many different aspects of libsai, in this case i allow to configure some dash api, like every variable in config.h and saidash.h keys, you can take a look of use case in syncd here: https://github.com/sonic-net/sonic-sairedis/blob/master/syncd/Syncd.cpp#L3165 when we load profile from file, if this config will be supproted in dash then we can modify easy those parameters from syncd pointing to right configuration files

kcudnik:
can we have this merged ?

kcudnik:
thanks!

PR #420, opened 2023-08-07
[SAI] Fix NextObjIndex method to embed objeect type into generated OID

kcudnik:
Couple changes:

* fix NextObjIndex and rename to NextObjectId, which embeds correct object type, this will be needed in sai_object_type_query
* fix dome log messages in get_switch_attribute
* move GetDeviceId to static, since no longer required to be public
* fix init_switch.cpp, since previously was using GetDeviceId to init GRPS channel, now  this is done inside sai_api_initialize

### DIFF
--- a/dash-pipeline/SAI/templates/utils.cpp.j2
+++ b/dash-pipeline/SAI/templates/utils.cpp.j2
@@ -111,7 +111,7 @@ bool InsertInTable(std::shared_ptr<p4::v1::TableEntry> entry, sai_object_type_t 
 
     if (*objId == 0)
     {
-        *objId = NextObjIndex(objectType);
+        *objId = NextObjectId(objectType);
     }
 
     tableEntryMap.insert(std::make_pair(*objId, entry));
@@ -119,7 +119,7 @@ bool InsertInTable(std::shared_ptr<p4::v1::TableEntry> entry, sai_object_type_t 
     return true;
 }
 
-sai_object_id_t NextObjIndex(sai_object_type_t objectType)
+sai_object_id_t NextObjectId(sai_object_type_t objectType)
 {
     DASH_LOG_ENTER();
 
--- a/dash-pipeline/SAI/templates/utils.cpp.j2
+++ b/dash-pipeline/SAI/templates/utils.cpp.j2
@@ -6,6 +6,8 @@ static std::unordered_multimap<sai_object_id_t, std::shared_ptr<p4::v1::TableEnt
 static std::mutex tableLock;
 static std::atomic<sai_object_id_t> nextId;
 
+#define MUTEX std::lock_guard<std::mutex> _lock(tableLock);
+
 std::unique_ptr<p4::v1::P4Runtime::Stub> stub;
 
 void correctIpPrefix(void *ip, const void *mask, size_t length)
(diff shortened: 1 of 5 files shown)


dash-pipeline/SAI/templates/utils.cpp.j2
```diff
@@ -94,38 +96,46 @@ grpc::StatusCode MutateTableEntry(std::shared_ptr<p4::v1::TableEntry> entry, p4:
...
+
     return true;
 }
 
-sai_object_id_t NextObjIndex()
+sai_object_id_t NextObjectId(sai_object_type_t objectType)
 {
     DASH_LOG_ENTER();
 
+    // TODO make proper oid
```
chrispsommers, line 126: Don't you want to address this now as well?

dash-pipeline/SAI/templates/utils.cpp.j2
```diff
@@ -94,38 +96,46 @@ grpc::StatusCode MutateTableEntry(std::shared_ptr<p4::v1::TableEntry> entry, p4:
...
+
     return ++nextId;
```
chrispsommers, line 128: If you're going to fix this, shouldn't there be a `nextId` per object type?

chrispsommers (changes requested):
I'm not sure of the goal of this PR, but it seems to leave the matter of maintaining nextId-per-object type unfinished. I've no objection to the changes so far though.

dash-pipeline/SAI/templates/utils.cpp.j2
kcudnik: i left this, since i didn't know whether tests will pass, but updated and fixed

kcudnik:
> I'm not sure of the goal of this PR, but it seems to leave the matter of maintaining nextId-per-object type unfinished. I've no objection to the changes so far though.

this is next pr in the series of refactoring libsai in dash rep, i don't want to put everything in one big pr, and then face looking for errors if something breaks :D

dash-pipeline/SAI/templates/utils.cpp.j2
chrispsommers: >i didn't know whether tests will pass

I looked at your fork, you do not have CI actions enabled. You can enable them and perform all CI testing *prior* to making a PR. This has several advantages - you can try uncertain things out without consequence (and without alerting/distracting watchers of the main repo), you can ensure your PR will pass before you even submit it, etc. I've advised all DASH contributors to do this. It runs in Azure cloud for free, same as DASH CI.

dash-pipeline/SAI/templates/utils.cpp.j2
kcudnik: addressed

dash-pipeline/SAI/templates/utils.cpp.j2
kcudnik: thats sounds good, how i can enable them ?

chrispsommers: Just go to your forked repo, select Settings gear on top ribbon, then click Actions on the left, then you'll see a dialog box where you can enable actions. 
gitactions

chrispsommers (approved):

PR #404, opened 2023-07-18
Adds dash underlay routing functionality 

AmithGspn:
**Description of PR**
This PR is the start for the underlay implementation in BMv2 of the design mentioned in the dash hld under swss lite
The focus of this PR is on introducing the implementation of SAI call for "SAI_OBJECT_TYPE_ROUTE_ENTRY". 

The keys implemented are "switch_id", "vr_id", "destination". 

The attributes implemented are  "SAI_ROUTE_ENTRY_ATTR_NEXT_HOP_ID". In this PR we have added support for next hop type "SAI_OBJECT_TYPE_PORT". Other types of next_hop_id can be implemented in subsequent PR's.

DASH-libsai code generator added support for fixed SAI headers.

It serves as a foundational step for future implementations, which aim to comprehensively cover the entire underlay scenario.

This PR lays the groundwork for further advancements in addressing the broader scope of the issue at hand.

Also it partially addresses https://github.com/sonic-net/DASH/issues/236 and test cases to the CI to ensure proper functionality

**What is the motivation for this PR?**
In summary, the motivation for this PR is to enable underlay support in BMv2 and implement the necessary SAI-Thrift calls to ensure the correct functionality of the routing table in the context of https://github.com/sonic-net/DASH/issues/236.

**Functionality**
The current implementation defaults to an "echo" behavior, where the packet is reflected back on the same port it arrived.

However, this behavior has been modified by incorporating a routing table. As a result, when there is no match in the routing table, the packet continues to be reflected on the same port.

In this PR, when a route is matched in the routing table, the packet is directed to a different port as specified by the matching route.

**Main Changes**
- Add Underlay routing table P4 logic for routing the packets based on match.
- Add VNET API for testing the functionality.
- Add the following test cases
  - Routing Unidirectional (Send packet from one port and receive the packet on other port upon route match)
  - Routing Bidirectional (Send packets from both ports of bmv2 switch and receive the packets on other ports upon route  
  match)

**How did you verify/test it?**
Can run the following test cases successfully:
- Previous test case showing the default no route behavior
- test_config_vnet_route_bidirectional.py
- test_config_vnet_route_unidirectional.py

PR #422, opened 2023-08-10
Adds SAI_ROUTE_ENTRY_ATTR_PACKET_ACTION api to dash_underlay_routing

AmithGspn:
**Description of PR**
This subsequent PR is the continuation of dash_underlay implementation PR #404  in BMv2 of the design mentioned in the dash hld under swss lite.

The focus of this PR is on introducing the implementation of SAI call for "SAI_ROUTE_ENTRY_ATTR_PACKET_ACTION" 

The attribute implemented is "SAI_ROUTE_ENTRY_ATTR_PACKET_ACTION". In this PR we have added support for packet action types "SAI_PACKET_ACTION_DROP" and "SAI_PACKET_ACTION_FORWARD". 

**Functionality**
In the current implementation, when there is no match in the routing table, the packet continues to be reflected on the same port and when a route is matched in the routing table, the packet is directed to a different port as specified by the matching route.

This PR decides whether to drop or forward the packet first. If the packet is to be forwarded, it goes through the next hop functionality else drops the packet.

**Main Changes**
- Add "packet_action" action to Underlay routing table P4 logic.
- Add VNET API for testing the functionality.

**How did you verify/test it?**
Can run the following test cases successfully:
- test_config_vnet_route_bidirectional.py
- test_config_vnet_route_unidirectional.py

### DIFF
--- a/dash-pipeline/bmv2/underlay.p4
+++ b/dash-pipeline/bmv2/underlay.p4
@@ -33,6 +33,16 @@ control underlay(
 #endif // TARGET_DPDK_PNA  
     }
 
+    action pkt_act(bit<9> packet_action, bit<9> next_hop_id) {
+        if(packet_action == 0) {
+            /* Drops the packet */
+            meta.dropped = true;
+        } else if (packet_action == 1) {
+            /* Forwards the packet on different/same port it arrived based on routing */
+            set_nhop(next_hop_id);
+        }
+    }
+
     action def_act() {
 #ifdef TARGET_BMV2_V1MODEL
         standard_metadata.egress_spec = standard_metadata.ingress_port;
@@ -61,8 +71,9 @@ control underlay(
         }
 
         actions = {
-            /* Send packet on different/same port it arrived based on routing */
-            set_nhop;
+            // Processes a packet based on the specified packet action.
+            // Depending on the packet action, it either drops the packet or forwards it to the specified next-hop. 
+            pkt_act;
 
             /* Send packet on same port it arrived (echo) by default */
             @defaultonly def_act;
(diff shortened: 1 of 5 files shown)


dash-pipeline/bmv2/underlay.p4
```diff
@@ -33,6 +33,16 @@ control underlay(
 #endif // TARGET_DPDK_PNA  
     }
 
+    action pkt_act(bit<9> packet_action, bit<9> next_hop_id) {
+        if(packet_action == 0) {
```
chrispsommers, line 37: Would prefer to use same symbol names as the C code i.e. `SAI_PACKET_ACTION_FORWARD, SAI_PACKET_ACTION_DROP` and `#define` them at the top of the file, with a comment referencing the sai header files they come from. (It's not practical to include the sai header files as-is into the P4 code).

chrispsommers (changes requested):
I see changes to test data (.json) but it's not clear which test cases read them and confirm both forwarding and drop actions. Could you please identify the test `.py` files involved? I would have expected some changes to Python code as well.

AmithGspn:
> I see changes to test data (.json) but it's not clear which test cases read them and confirm both forwarding and drop actions. Could you please identify the test `.py` files involved? I would have expected some changes to Python code as well.


No modifications are required within the Python files. These Python scripts are solely responsible for creating and removing APIs, as well as generating packets from designated ports. It is important to note that no code alterations associated with this pull request are contingent upon changes within these Python files.

dash-pipeline/bmv2/underlay.p4
AmithGspn: yea working on it.

chrispsommers (approved):

PR #423, opened 2023-08-10
[SAI] Refactor SAI and add new features to be compatible with sonic-sairedis 

kcudnik:
Many changes:

* Move utils to it's own namespace dash::utils 
* Move fixed api implementation to proper DashSai class
* Remove unused methods
* Add workaround to return not supported status on get api 
* Remove underscore from variable naming
* Fix route bulk api set/get warnings
* Simplify insertInTable method
* Change asserts to return error codes and add missing switch api
* Print oids in hex form

### DIFF
--- a/dash-pipeline/SAI/templates/dashsai.cpp.j2
+++ b/dash-pipeline/SAI/templates/dashsai.cpp.j2
@@ -262,7 +262,7 @@ sai_status_t DashSai::createSwitch(
 
     *switch_id = m_switchId;
 
-    DASH_LOG_NOTICE("created switch id: %ld", *switch_id);
+    DASH_LOG_NOTICE("created switch id: 0x%lx", *switch_id);
 
     return SAI_STATUS_SUCCESS;
 }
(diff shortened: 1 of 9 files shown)


kcudnik:
it should be better to review one commit at a time

dash-pipeline/SAI/templates/dashsai.cpp.j2
```diff
@@ -0,0 +1,649 @@
...
+ *
+ * Name of environment variable, when set to any value, will return proper
+ * SAI_STATUS_NOT_SUPPORTED on switch/port GET api.
+ *
+ * Currently because of limitation of saithrift, GET on switch/port is
+ * returning SAI_STATUS_SUCCESS on not implemented attribute.
+ *
+ * TODO FIXME needs to be addressed in saithrift.
+ *
+ * This is temporary workaround for syncd, it should be removed when fixed in saithrift.
```
chrispsommers, line 28: Is there a corresponding issue filed against saithrift?

dash-pipeline/SAI/templates/saiapi.cpp.j2
```diff
@@ -509,6 +513,35 @@ static sai_status_t dash_sai_get_{{ table.name }}_attribute(
     assert(0 && "sai_get_{{ table.name }}_attribute NYI");
     return SAI_STATUS_FAILURE;
 }
+
+{% if table.name == 'route_entry' %}
```
chrispsommers, line 517: Why this particular case?

chrispsommers (approved):
LGTM

dash-pipeline/SAI/templates/dashsai.cpp.j2
kcudnik: not yet

Issue #395, opened 2023-06-25
SAI Meta out of libsai


KrisNey-MSFT:
Per June 21 2023 DASH Community Call.

Should we compile SAI Meta into SAI Thrift?  This is super old, not sure how to answer how it goes together.  There might be missing symbols.  SAI Meta out of libsai would be the Issue - K to create & assign (Marian, Chris, Vijay, K~)

KrisNey-MSFT:
@vijasrin to further articulate details of the ask

mukeshmv:
Ask is to be able to run DASH SAI thrift tests on the same vendor SAI library that is used in the SONiC production system with SONiC syncd. 
The problem is that DASH SAI thrift server fails to build unless the vendor SAI library is linked with the SAI meta since we run into undefined reference for the following symbol otherwise: sai_metadata_get_attr_metadata.

However, in the case of production system, the SONiC syncd links the libsaimetadata, so there is no need for the vendor SAI library to include the SAI meta objects again. This makes sense since the SAI meta info is generated based on the SAI API definition and hence constant across all vendors, so it does not belong in the vendor SAI library.

chrispsommers:
I'll take a look!

chrispsommers:
I made a temporary branch of DASH here: https://github.com/chrispsommers/DASH/tree/no-saimeta-in-libsai. This removes sai-meta from bmv2 libsai, but a problem remains.

If you `make clean && make all` you get unresolved symbol errors while building saithrift-server. This might be an oversight in the saithrift-server build recipe itself. It assumes that the vendor libsai contains saimeta utils, which is why DASH bmv2 libsai had code to build certain files, e.g. see https://github.com/sonic-net/DASH/blob/64cbb0ae83f65b14469f8554ace942ca18c3a8d1/dash-pipeline/SAI/templates/Makefile.j2#L5 and related compile instructions. (I'd arrived at all this empirically long ago, without questioning the SAI saithrift-server build process.)

So, with this modified branch, sai-meta code is not compiled into bmv2 libsai. If you do the following:
```
make clean && make all
```
it will build p4 code, build sai-headers, (*not* build sai-meta anymore), build libsai - all OK up to this point - then fail to build saithrift-server during the final linkage stage, as follows:
```
g++ -L/usr/lib -Wl,-rpath=/usr/lib ./obj/sai_rpc_server.o ./obj/saiserver.o -o saiserver \
[...]

PR #431, opened 2023-09-01
[submodule] Update SAI submodule to origin/master

kcudnik:
To include
[test] Refactor saithriftv2 Makefile
https://github.com/opencomputeproject/SAI/commit/4f39bd


kcudnik:
@chrispsommers please review and if good to go please merge

chrispsommers (approved):
LGTM, thanks!

chrispsommers:
@kcudnik Does this pave the way to resolve https://github.com/sonic-net/DASH/issues/395 ? In particular, does sai-thrift still require sai meta code inside vendor libsai?

kcudnik:
> @kcudnik Does this pave the way to resolve #395 ? In particular, does sai-thrift still require sai meta code inside vendor libsai?

No, currently saimetadata*.o objects are compiled into libsai.so (from dash)

for example: in ~/SAI/test/saithriftv2/Makefile you can se saimetadata lib added when saivs is compiled
```make
 40 ifeq ($(platform),vs)
 41 LIBS = -lthrift -lpthread -lsaivs -lsaimeta -lsaimetadata -lzmq
 42 else
 43 LIBS = -lthrift -lpthread -lsai
 44 endif
```

so currently that libsai.so from DASH trhift will require that metadata, i have PR almost ready that will simplify this, Makefile DASH/dash-pipeline/SAI/templates/Makefile.j2, and metadata function sai_metadata_get_attr_metadata is required from SAI/meta/sai_rpc_frontend.cpp, since they need to get a attribute type to properly convert sai_attribute to sai_thrift_attribute, and this is manually created file, not sure if any other things could be used in the future in rpc frontend

currently im compiling saimetadata into libsai.so to make it link correctly, but if it's planned to use other libsai than generated from DASH, then we can update SAI/SAI/test/saithriftv2/Makefile to store those metadata objects in that librpcserver.a archive

kcudnik:
since many different parties are using or will use some part of saimetadata, i propose to make libsaimetadata.so and share it everywhere instead to linking objects in each project, deal later with duplicate references, and make size increase of each file, libsaimetadata.so can be already generated by SAI repo latest commit :)

please take a look at further improvements: https://github.com/sonic-net/DASH/pull/432

PR #437, opened 2023-09-11
[dash-pipeline] Refactor docker files

kcudnik:
For more readability


dash-pipeline/dockerfiles/Dockerfile.saithrift-client
```diff
@@ -10,23 +10,22 @@ the saithrift client and test programs for DASH."
...
-    cd thrift-0.11.0 && \
-    sudo python3 setup.py install && \
-    cd / &&\
-    sudo rm -rf thrift-0.11.0 && \
-    cd /SAI/test/ptf && \
-    sudo python3 setup.py install
+RUN cd /saithrift-0.9 && sudo python3 setup.py install
+RUN sudo rm -rf /saithrift-0.9
+RUN cd thrift-0.11.0 && sudo python3 setup.py install
+RUN sudo rm -rf /thrift-0.11.0
```
chrispsommers, line 27: Please see https://stackoverflow.com/questions/39223249/multiple-run-vs-single-chained-run-in-dockerfile-which-is-better for a discussion on the tradeoffs of one RUN command vs several. In summary, multiple RUN commands can improve efficiency of shared layers, assuming any are shareable (generally not true in our project). It's arguably prettier to read. However, having `rm` commands in a separate layer can result in downloading the prior layer containing the content you want to remove, then having the layer containing `rm` deleting it from the filesystem, *after* you've already downloaded it from the layer before, so it's inefficient.

kcudnik: i added this just for readability

chrispsommers: Understood, and I appreciate the effort, but we need to follow a few best practices for layers.

PR #443, opened 2023-09-21
update p4 compile dependency to avoid parallel identical docker runs for both the json and the p4rt.txt

vincent-xs:
p4 depends on 
P4_ARTIFACTS=$(P4_OUTDIR)/dash_pipeline.json $(P4_OUTDIR)/dash_pipeline_p4rt.txt

the makefile was modified to include a P4_ARTIFACTS rule. this means that 2 identical parallel docker runs will be launched (one for each of the files) causing an issue when compiling p4 (DOCKER_FLAGS= make -j p4)

the fix is to create a rule for the json file (to depend on p4rt) and update the docker run rule to be for the p4rt file.

### DIFF
--- a/dash-pipeline/Makefile
+++ b/dash-pipeline/Makefile
@@ -103,8 +103,10 @@ p4: $(P4_ARTIFACTS)
 p4-clean:
 	rm -rf $(P4_OUTDIR)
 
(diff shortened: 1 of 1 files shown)


chrispsommers:
Thanks Vincent!

PR #448, opened 2023-10-03
Add per ENI DSCP setting

marian-pritsak:
By default, use DSCP from the original outer packet.


dash-pipeline/bmv2/dash_metadata.p4
```diff
@@ -28,11 +28,18 @@ struct conntrack_data_t {
     bool allow_out;
 }
 
+enum bit<16> dash_tunnel_dscp_mode_t {
+    INVALID = 0,
+    PRESERVE_MODEL = 1,
```
r12f, line 33: should we make one of preserve_model or explicit set as default instead of invalid?

r12f: also I guess you mean `PRESERVE_MODE` : D

dash-pipeline/bmv2/dash_metadata.p4
```diff
@@ -28,11 +28,18 @@ struct conntrack_data_t {
...
+    PIPE_MODEL = 2
```
r12f, line 34: somehow I feel `PIPE_MODE` might not be that intuitive.... but will `EXPLICIT_MODE` be a better name? not sure if it is more clear... but no strong opinion.

dash-pipeline/bmv2/dash_metadata.p4
r12f: Actually, please ignore my comment, I think what you are doing is right.

dash-pipeline/bmv2/dash_metadata.p4
r12f: Please ignore my comment above. What you are doing should be right.

dash-pipeline/bmv2/dash_pipeline.p4
```diff
@@ -116,6 +116,8 @@ control dash_ingress(
                          bit<16> vnet_id,
                          bit<16> v4_meter_policy_id,
                          bit<16> v6_meter_policy_id,
+                         dash_tunnel_dscp_mode_t dash_tunnel_dscp_mode,
```
r12f, line 119: feels like we can simply call it `tunnel_dscp_mode`

r12f: and to make things match the dscp field can be `tunnel_dscp_value`.

KrisNey-MSFT:
hi @marian-pritsak - is this ready for a Merge?  Let me know :)

Issue #428, opened 2023-08-28
Clarification on Service Tunnel Inbound Processing


budgrise:
The HLD describes service tunnel inbound processing:
https://github.com/sonic-net/DASH/blob/main/documentation/general/dash-sonic-hld.md

I have a few questions on the requirements.

In Section 2.3 "Service Tunnel (ST) and Private Link (PL) packet processing pipelines":
The text says: "ST/PL Inbound flow: Using the outbound unified flow, the reverse transposition (inbound unified flow) is created."

It was not clear to me what this is stating. Is it:
A) The outbound packet makes its way to the remote VM, and then the packet returned from the VM has its innermost SIP/DIP swapped.
B) Or, during processing of the outbound packet, the appliance creates a flow table entry for the reverse flow. And since the incoming outbound packet's innermost header is IPv4, and the incoming inbound packet's innermost header is IPv6, outbound processing creates two flow table entries?
- an IPv4 entry for matching on subsequent outbound packets 
- an IPv6 entry for the inbound packets

In Section 3.6.2 "Service Tunnel":

1)  I believe the description says the packet sent to the appliance is double-tunneled, is this correct? 
e.g. the header stack looks like:
	ETH IPV4 NVGRE  ETH IPV4 NVGRE  ETH IPV6 TCP

2a) h.4 says "Appliance shall first decapsulate the outer header and map it to a flow."
Does this mean decapsulate the first outer header or both the outer and middle headers?

2b) The flow table keys include IP addresses and L4 ports. What header fields from this stack are used for the flow table lookup?

budgrise:
@prsunny  you might be the best to answer these.

KrisNey-MSFT:
pinged @prsunny re: this after Community Call 8/30/2023

KrisNey-MSFT:
2 flow entries, 1 IPv4 and 1 IPv6
How to handle double encap (2 headers) before the flow lookup - what is the logic?
Someone create a 'comment' or suggestion in the document location in SDN Packet Transforms or in SONiC HLD?

KrisNey-MSFT:
Close the PRs, create Issues instead

r12f:
yes, in the ST case, the outbound packet will evaluates the DASH policies and eventually create 2 flows:
- Outbound direction forwarding flow: IPv4 flow, matching the IPs from the outbound side innermost IPv4 packet
[...]

Issue #406, opened 2023-07-24
nvgre parsing


budgrise:
Nvgre output was added by PR #308 Add ST route type.
Is there also a requirement for parsing incoming nvgre packets?

KrisNey-MSFT:
No support for packets coming to the pipeline, they can only go out.  
No use for it right now (no implementation for nvgre on ingress), but may be in the future, and we would need to add the code.  
Assumption is that 2 DASH devices would not communicate using nvgre.  
ST is a tunnel towards a public IP (on the Internet).  Do packets return on nvgre?  Ask @prsunny

KrisNey-MSFT:
Added @pranjal-git in case he may know

r12f:
Hi @budgrise, on high level, the packet transformation will still follow this doc: <https://github.com/sonic-net/DASH/blob/main/documentation/general/sdn-pipeline-basic-elements.md#service-tunneling>, and I am sure you are already aware of it:

In my understanding, the return packet will rely on reverse flow to do the transformation:

- The return packet will have 2 layers of encaps, which will be all decap'ed before sending to the pipeline. However, this is not modeled today in DASH and our BM, but this is what should be happening.
- After decap, the flow matching will only use the 5 tuples from the inner most packet. This will find the reverse flow and apply the transformations.

r12f:
Hi Bud, I have updated the proposal for making this behavior clarified as well as addressing certain future cases. Please feel free to take a look and let me know what you think: https://github.com/r12f/DASH/blob/user/r12f/pipeline/documentation/general/dash-sai-pipeline-packet-flow.md#54-packet-decap.

The PR can also be found here: https://github.com/sonic-net/DASH/pull/449. Feel free to leave me comments as well! :D

KrisNey-MSFT:
Closing in favor of PR449 per Community Call

Issue #386, opened 2023-06-07
Ensure we can decap double-encapsulated packets - Add to DASH HLD Section 2.3


KrisNey-MSFT:
The source side is covered in DASH.  ST will always have 2 encaps.  Reply will have 1 encap.  If traffic is cross-region, then ST gwy then another layer of encap.  SRC mux is not involved.  Reply packet goes to SRC SLB Mux for another layer of encap.  

ILB on DST side is n**ot covered**, however will be double encapp’d.  
1 VNET encap, 1 SLB encap.

This is an open item in DASH on the DST side.  We need to model this in DASH and add to HLD here:

https://github.com/sonic-net/DASH/blob/main/documentation/general/dash-sonic-hld.md#23-service-tunnel-st-and-private-link-pl-packet-processing-pipelines

Issue #117, opened 2022-06-01
NVGRE or VxLAN or both


mgheorghe:
Hero test talks about VxLAN as tunneling protocol 
https://github.com/Azure/DASH/blob/main/documentation/general/requirements/program-scale-testing-requirements-draft.md

some design documents show NVGRE as tunneling protocol
https://github.com/Azure/DASH/blob/main/documentation/general/design/images/service_tunneling.png

what will be supported by Dash ? NVGRE, VXLAN, or both ?
any other tunneling protocol to be supported ?

and to what extent, will full VxLAN and full NVGRE be supported as per spec?
https://datatracker.ietf.org/doc/html/rfc7348
https://datatracker.ietf.org/doc/html/rfc7637
....

PR #449, opened 2023-10-05
DASH pipeline packet flow update proposal.

r12f:
# Motivation

DASH pipeline packet flow is the core for data path modeling, and today, it provides a good starting point for modeling VNET traffic. However, certain behaviors of DASH pipeline is not well defined or not being flexible enough.

For example:

- Only certain VM related scenarios are supported. 
  - Inbound pipeline is not fully modeled.
  - Things like load balancer will require large change to today's pipeline, e.g. PR 348: <https://github.com/sonic-net/DASH/pull/348>.
- Routing types being the minimum building block makes it hard to add any new scenarios, because we have to change our behavior model for every new scenarios. This is not scalable for everyone that involved in DASH - both users or vendor implementations. E.g.:
  - PR 327: <https://github.com/sonic-net/DASH/pull/327>
  - PR 273: <https://github.com/sonic-net/DASH/pull/273>
- Certain routing actions behaves differently under different routing types.
  - “staticencap” uses “underlay_ip” from ENI in regular scenario, while “pl_underlay_ip” in PL scenario.
  - “4to6” action uses pl_sip_encoding field from ENI to encode the overlay_sip, while in ST, it uses only overlay_sip.
- DASH object contains fields from Microsoft only scenarios, such as private link.

And this PR is to:
- Provide the doc to have the packet flow more well defined to address the unclear part of the pipeline.
  - #406
  - #428
  - #386
  - #448 
  - #117
- Propose some extension to today's DASH pipeline to make it more flexiable, so more scenarios can be onboarded without changing the underneath behavior model.

# Change summary

This proposal is not indented to redesign the SAI API and DASH packet flow today, but mostly focusing on a few extensions in today's DASH model.

- Exposing routing type table and make routing actions as the new building block.
  - To transit to the model, we could have them both co-exist side by side, sharing the same name. Until we are ready to switched over.
- Simplify the pipeline as entry matching and metadata publishing and removing majority of the special logics for any routing actions.
- Adding port mapping stage to the existing pipeline and model the inbound pipeline the same way as outbound.


documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,799 @@
...
+
+A pipeline can also define its initial matching stages, which will be used for starting processing the packets if no flow is matched.
+
+> NOTE:
+> 
+> DASH-SAI pipeline is a logical concept. It doesn't have to be 1 to 1 mapping to a physical ASIC pipeline. The underlying implementation can be as simple as a metadata field update with a specific value, when the entry getting matched, e.g., `ENI = Inner Source/Destination MAC`.
+
+### 5.4. Packet Decap
+
+If a pipeline is found, before processing the packets, all outer encaps will be decap'ed, and with key information saved in metadata bus, such as encap type, source IP and VNI / GRE Key, exposing the inner most packet going through the pipeline. This simplifies the flow matching logic and also allow us to create the reverse flow properly.
```
budgrise, line 212: Can you clarify how to decap all outer encaps? I'm thinking we probably don't want to simply parse headers until there are no more tunnels found. What if the customer traffic is actually a tunnel they created?

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,799 @@
...
+
+After entering a specific pipeline, the first stage will be the Conntrack Lookup stage, which does the flow lookup. If any flow is matched, the saved actions will be applied, the metering counters will be updated, and the rest of pipeline will be skipped.
+
+The core of the Conntrack Lookup and Update stage is the flow table.
+
+#### 5.5.1. Flow lookup
+
+First, let's define the flow lookup behavior.
+
+The flow lookup **MUST** use the information of inner most packet. And the matching keys can be configurated via the DASH flow APIs during the DASH initialization. This allows us to support different flow matching behaviors for certain cases. For example, in L3 level routing, we don't need to create 5 tuple flows.
```
budgrise, line 237: If I understand correctly, you are not saying that arbitrary fields can be used as the flow keys. Instead, from the packet there is at most the classic 5-tuple fields. And any of these 5 fields can be disabled from the key based on a config given at DASH init time? (To disable the field basically means using a fixed value like 0).

r12f: Hi Bud, thanks a lot for review and leaving comments! For your questions:

> If I understand correctly, you are not saying that arbitrary fields can be used as the flow keys. Instead, from the packet there is at most the classic 5-tuple fields.

Yes, modeling the arbitrary fields as flow key will be very complicated, and also hard to implement on certain DPU/ASICs. So, the classic 5 tuple besides certain frequently used fields should be a place to start with. More details can be found here in the flow api spec (in review): See `sai_dash_flow_key_t` in <https://github.com/sonic-net/SONiC/blob/d2bb1913dacf682cfa2affc0723101df5fc6ed81/doc/dash/dash-flow/DASH_FLOW_SAI.md>. 

> And any of these 5 fields can be disabled from the key based on a config given at DASH init time? (To disable the field basically means using a fixed value like 0).

Yes, and yes, using 0 is actually a valid way to implement this. There are certain DPUs that actually allow the stages to be dynamically created too, so the underlying implementation is flexible, depending on the architecture.

documentation/general/dash-sai-pipeline-packet-flow.md
r12f: yea, I was thinking about adding these details but also afraid going too far on this XD. But here is my thoughts:

## Problem of preserving the encaps

Today, the encaps of incoming packet (before pipeline processing) will be used for creating the encaps of the reverse flow, so essentially, we have 2 things to answer:

1. How preserving encaps affects the reverse flow creation?
2. How preserving encaps affects the forwarding side tunnel creation?

With the packet flow doc, now we have a clear definition of the packet structure: overlay (inner most) -> underlay -> tunnel0 -> tunnel1 -> .... However, depends on what user wants, they could do something really interesting, for example, removing or updating the underlay, preserve the tunnel0 and remove tunnel1. So, we are facing all the combinations of CURD of all the encaps including structure change: after removing underlay, should tunnel0 becomes underlay or should we keep it as tunnel0? This affects how the `staticencap` and `tunnel` action works today.

## Solution

In my thought, since these are essentially 2 questions, then we can model it as 2 independent operations. 

### Reverse tunnel creation

For reverse tunnel creation, Since no matter which encap (nvgre or vxlan) we use, they always have a key inside, so we can build the VNI table to specify what we should do. This is already covered with "stateless decap and stateful decap" section. If a VNI is stateless, then we don't create the reverse encap, otherwise we create one.

```json
{
    "DASH_SAI_VNI_TABLE|12345": { "direction": "outbound", "stateless": false },
    "DASH_SAI_VNI_TABLE|12346": { "direction": "outbound", "stateless": true }
}
```

If automatic reverse tunnel creation is not enough, we can also use `reverse_tunnel` to create a dedicated one.

### Forwarding tunnel creation

Although the encap is decaped, but all the encap information is preserved in the metadata bus. Hence, we can easily recreate them in anyway we want by introducing a new action, say: `tunnel_from_encap` (sorry, the name might look silly), and defined as below:

- Parameters from action:
  - Source: "underlay|tunnel0|tunnel1|..."
  - Target: "underlay|tunnel0|tunnel1|..."
- Parameters from metadata:
  - `tunnel_from_encap_(underlay|tunnel0|tunnel1)_sip`: overrides the source ip in the specified target encap
  - `tunnel_from_encap_(underlay|tunnel0|tunnel1)_dip`: overrides the destination ip in the specified target encap
- Action:
  - Enable the target encap by copying the information from the source encap.
  - When overrides exists, uses the overrided one.

If this makes sense, I will add them into the doc.

documentation/dataplane/dash-routing-actions.md
```diff
@@ -0,0 +1,197 @@
...
+
+- `use_src_ip`: Use source IP in routing and mapping stages.
+- `ip_mask`: IP mask to apply before matching the entries.
+
+##### `lpmrouting` action
+
+- Actions:
+  - The packet will be routed to the next stage based on the LPM routing table.
+
+The port mapping entries can be described as below:
```
ashutosh-agrawal, line 69: Typo. You probably want to mention lpmrouting entries instead of "port mapping" entries.

documentation/dataplane/dash-routing-actions.md
```diff
@@ -0,0 +1,197 @@
...
+
+- Actions:
+  - Trap the packet to CPU after the entry is matched.
+
+### Matching stage transition actions
+
+All matching stage transition actions will support the following parameters:
+
+- `default_routing_type`: If no entry is found, use this routing type to route the packet. If `default_routing_type` is not set, the packet will be dropped by default.
+- `stage_index`: If there are multiple stage of the same stage type, use this field to specify which stage to transit to. The index is starting from 0 and by default 0.
```
ashutosh-agrawal, line 55: Can you please elaborate on multiple stage requirements a bit more. Today, the pipeline has a single routing and mapping stage along with multiple ACL stages. If there can be multiple routing stages, how is the action resolution expected to happen? Would there be a priority attached to each stage or the later stage would override the conflicting action from an earlier stage?

documentation/dataplane/dash-routing-actions.md
```diff
@@ -0,0 +1,197 @@
...
+
+- Metadata parameters:
+  - `underlay_dip`: Destination IP used in encap.
+  - `underlay_sip`: Source IP used in encap.
+  - `encap_type`: "nvgre|vxlan|…"
+  - `encap_key`: GRE key in NvGRE or VNI in VxLAN
+- Actions:
+  - Enable the underlay encap header based on the encap type.
+  - Update the underlay encap header with `encap_key`, `underlay_dip`, `underlay_sip`.
+    - If `underlay_dip` / `underlay_sip` is not set, use the original IP.
```
ashutosh-agrawal, line 139: If we are encapsulating a tenant packet and let's say only underlay_dip is set in the action, then are we supposed to copy tenant(overlay) sip to underlay_sip? Then packet will go out with an encap consisting of dip from the underlay address range and sip from the overlay address range. Doesn't sound right to me.

documentation/dataplane/dash-routing-actions.md
r12f: Right, it is a typo. I will rephase this sentence and update again.

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,799 @@
...
+- If even more customized or complicated pipelines are needed for certain traffic, we could also provide our own customized data plane app and use the DASH policy to instruct the ASIC to trap these packets up when they shows up.
+
+## 3. Packet Structure
+
+Before diving into the pipeline and packet flow, to better describe the behaviors, let's first have a clear understanding on the packet structure that DASH supports.
+
+Overall, the high-level packet structure looks like below:
+
+- The inner most packet is whatever the customer sends, which is called overlay. 
+- The first encap is called underlay, which is the most frequently used layer for implement any virtual network functions, such as VNET routing, load balancer, etc.
```
ashutosh-agrawal, line 66: Since the underlay also requires a tunnel encapsulation, wouldn't it be simple to model the layers as | ** | Tunnel 2 | Tunnel 1 | Tunnel 0 | Overlay | where Tunnel0 is used for underlay?

r12f: that's actually true. however, this underlay concepts is already everywhere in today's DASH pipeline, such as mapping table, eni and etc.... so maybe better to keep it as it is for now.

but to allow this change for future, we could use | ** | Tunnel 2 | Tunnel 1 | Underlay | Overlay |. does this sound good to you?

r12f: updated.

documentation/dataplane/dash-routing-actions.md
r12f: fixed.

documentation/general/dash-sai-pipeline-packet-flow.md
r12f: Updated the doc according. Please feel free to take another look, Bud! : D

documentation/general/dash-sai-pipeline-packet-flow.md
ashutosh-agrawal: Yeah, that's much better. Now, Tunne0 and underlay can be used interchangeably, if needed.

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,839 @@
...
+- The information from the original packet, such as encap information.
+- The information from each matched entry and related data structures, e.g., when a VNET mapping entry is matched, we will publish the information from VNET to the metadata bus. 
+- The routing action and its inline parameters from the matched entry.
+
+Implementation-wise, this is similar to Packet Header Vector or Bus in NPL.
+
+### 5.2. Direction Lookup
+
+In DASH-SAI pipeline, traffic are split into 2 directions: `inbound` and `outbound`. Each direction has its own pipeline (see pipeline overview above). When a new packet arrives, we will assign a direction to the packet, then process the packet in the corresponding pipeline. This ensures us to match the flow and transform the packet in the right way.
+
```
ashutosh-agrawal, line 174: For completeness, it would be nice to specify how the direction is determined. Is it always based on the VNI lookup on the 1st VXLAN header found? Can we assume that every packet will have one and only one vxlan header? I believe some of the DASH use cases have multiple NVGRE headers, but I don't remember seeing a header stack with multiple vxlan headers.

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,839 @@
...
+### 5.3. Pipeline Lookup
+
+DASH supports multi-tenancy model for traffic handling. A single device can have multiple pipelines, and each pipeline is used to handle traffic for a specific tenant. When a packet arrives, besides direction lookup, we also need pipeline lookup to determine which pipeline to use for processing the packet.
```
ashutosh-agrawal, line 177: Nit: You may want to mention here that "each is pipeline is used to handle traffic for a specific tenant OR instance of a service (L4LB)"

documentation/dataplane/dash-routing-actions.md
r12f: It is actually similar to what you are thinking, and the process is very simple. The details are captured in the "Action Publishing" and "Metadata Publishing" sections, which covers the action itself and action parameters.

Say, we have 2 stage of LPM stage, and the process will be mostly like this:

- First, we create the LPM entries in the 2 LPM stages:
  - The stage index can be set by an attribute. Essentially, we can think it is LPM0 and LPM1.
  - Entry0 in LPM0 will have transition set to `lpmrouting` with `stage_index` set to 1, then when it matches, the second LPM stage will be invoked.
  - Entry1 in LPM1 can have the `routing_type` set, just like today and points to a list of actions, say: `staticencap`.
- Then, when a packet shows up, it matches Entry0, 
  - Transition info will be populated to metadata and make the pipeline skipping the internal stages (if any), and execute the second LPM.
  - Other metadata will also be populated, say, `underlay_dip` and etc.
- Then, the packet matches Entry0 in LPM1,
  - Transition info is not set, so we are reaching the end of matching stages.
  - Routing type is set, and `staticencap` action will be populated into metadata.
- When reaching ActionApply stage, the staticencap action will be executed, using the `underlay_dip` from the metadata bus.

More details and pseudo code can be found in "Stage skipping" and "Multi-stage chainning" sections.

documentation/dataplane/dash-routing-actions.md
r12f: The pipeline has the `underlay_ip` field, which will be published as `underlay_sip` in outbound pipeline and `underlay_dip` as inbound pipeline. 

I forgot to removed the last item and it is not removed to avoid confusion.

documentation/general/dash-sai-pipeline-packet-flow.md
r12f: Yes, multiple encaps could happen and we should be able to handle that. How to handle the multiple encap is already added in the "Packet Decap" section.

documentation/general/dash-sai-pipeline-packet-flow.md
r12f: right! this is actually a good one. Since LB is also just an example, I have updated this part to " or whichever network function being modeled, such as a load balancer".

r12f: updated now.

documentation/general/dash-sai-pipeline-packet-flow.md
budgrise: Hi Riff, you wrote "Today, the encaps of incoming packet (before pipeline processing) will be used for creating the encaps of the reverse flow".  I'm not sure if this is true with the dash behavior that's been defined so far. 

In regard to the double decap, I think you are saying run the pipeline twice? After the first tunnel header is decapped, run the pipeline and the LPM/actions etc will direct us to run the pipeline again (and thus trigger another decap)? If I have this wrong please correct me.

I have a higher-level comment sort of related to this and the rest of the doc. I see you are defining the dash pipeline to make it more generic and modular. But this is not the approach that has been taken up until now. Maybe the use cases for DASH have expanded? I'm a little bit concerned about the effort it will take to implement what you're proposing/defining.

r12f: > Hi Riff, you wrote "Today, the encaps of incoming packet (before pipeline processing) will be used for creating the encaps of the reverse flow". I'm not sure if this is true with the dash behavior that's been defined so far.

Hi Bud, feel free to correct me, if my understanding is not right. 

I believe what you are referring to is this: "For the inbound direction, after LPM/ACL lookup, the pipeline uses the underlay_ip as specified in the ENI table to Vxlan encapsulate the packet.". 

In my understanding, this is right when no flow is created on the inbound side. But if we allow outbound direction creating a flow and let the flow to handle the inbound packet, then this underlay_ip will not be used anymore, since the transformation should be coming from the flow. This maps to how fast path in DPU/ASIC. So, we need to get the IP and stored it into the reverse flow (for inbound side).

The simple idea is to use the underlay_ip in reverse flow. But if source VM using multiple VTEP as ECMP group, it won't be handled well, because it is preferred to use whichever VTEP that sends the traffic to return the packet. Additionally, if the ENI is used for modeling other things which handles multiple sources, such as load balancer or other gateways, the source IP will not be a single IP or can be configured on the ENI. So it is better for us to learn it from the original packet.

> In regard to the double decap, I think you are saying run the pipeline twice? After the first tunnel header is decapped, run the pipeline and the LPM/actions etc will direct us to run the pipeline again (and thus trigger another decap)? If I have this wrong please correct me.

Not really. We don't recirculate the packet, but let the parser to handle certain level of the encaps. What you are saying is actually a very good one. I believe the confusion is caused by we didn't have the maximum number of encaps defined. If this solves your concern, I will update the doc.

> I have a higher-level comment sort of related to this and the rest of the doc. I see you are defining the dash pipeline to make it more generic and modular. But this is not the approach that has been taken up until now. Maybe the use cases for DASH have expanded? I'm a little bit concerned about the effort it will take to implement what you're proposing/defining.

What you are feeling is right. I think DASH is doing good job modeling certain VNET scenarios. And it has all the corner stones to be flexible and modular, but it is not using them at all. For example, routing action behavior changes based on routing types. Routing types is used as the real minimum building block, making routing action almost useless. These things limit the usage of DASH a lot.

If you look at the proposal and ignore the inbound pipeline and multi-stage update, you will see the object model is not changed much at all (VNI table is introduced for handling multi-layer encap and multi-VNI cases, so it doesn't really count). But you might already feel the pipeline being much more flexible and can do a lot of more things.

For your concern, yes, I share the same concern as well. There are certain fundamental things being changed in this proposal, such as moving from programming routing type to programming routing action. It could increase the underlying implementation complexity. But on the other hand, with a simple mind model, the complexity could be highly reduced as well. No more corner cases, no more special handling, everything just moves as you see it.

So, although I have the same concern, I still would like to get this doc and proposal out, discussed and gather the feedbacks. And after this, we will be able to have a clear definition on how the pipeline works instead of having many actions being vague and do whichever they like under different conditions.

Hope this solves your question!

r12f: the maximum number of decap is added into the doc.

budgrise: Hi Riff,
Just writing down my comment from the last meeting regarding handling packets that arrive with two tunnel encaps. IMO always decapping the two tunnel headers is not a good solution. If the outer tunnel header is for DASH and the inner tunnel is a customer tunnel, then we would probably drop the packet. (The customer tunnel wouldn't match any valid DASH tunnel.) This basically means that DASH doesn't let customers use tunnels, which is kind of a severe limitation.

Not sure what the solution is. I think ideally you'd like to check the outer tunnel to know if a second tunnel is expected. For instance, if a mux can put a second encap on on a packet, then maybe the VNI/tunnel_id the mux applies is known to DASH orchestration. And the DASH direction_lookup table (which has the tunnel_id for a key) includes a double_encap flag. This would be set to true for any tunnel from a mux. At packet forwarding time, we do the direction_table lookup. If this flag is not set in the lookup result then we decap only the outer tunnel. If the flag is set, then not only do we decap a second tunnel, but we drop the packet if a second tunnel is not present. I know there are issues with implementing this with the P4 parser -- I'm only suggesting a logical solution.

r12f: Thanks a lot for capturing this Bud! I have updated the doc with multi-layer encap handling for this problem, which should be doable with P4 with `value_set` feature in parser. Do you mind to help take another look and see how it looks? : D

On high level, it works with 2 things:

1. use `final_encap` flag in VNI, so we can tell the parser that there is no more encap after this.
2. use some code to fix or extract the encap info in metadata after parsing, similar to what DASH is doing today.

KrisNey-MSFT (commented):
I made a few suggestions

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,886 @@
...
+
+| ... (Outer most) | 2 | 1 | 0 (Inner most) |
+| - | - | - | - | - |
+| **...** | **Underlay 1** | **Underlay 0** | **Overlay** |
+
+Today, DASH supports up to 2 layers of encaps: underlay0 and underlay1.
+
+## 4. Pipeline Overview
+
+DASH-SAI pipeline is modeled as a list of stages. Each stage defines its own tables, and use the table entries to match packets per the processing logic inside the match-action pipeline (key, action id, action data), and publish the corresponding metadata when an entry is matched. After all stages are processed, a list of final routing actions will be defined. Then, by executing these routing actions, the packet will be transformed accordingly, and corresponding flows will be generated for to the direction of the packet.
```
KrisNey-MSFT, line 85:
```suggestion
DASH-SAI pipeline is modeled as a list of stages. Each stage defines its own tables, and use the table entries to match packets per the processing logic inside the match-action pipeline (key, action id, action data), and publish the corresponding metadata when an entry is matched. After all stages are processed, a list of final routing actions will be defined. Then, by executing these routing actions, the packet will be transformed accordingly, and corresponding flows will be generated for the direction of the packet.
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,886 @@
...
+- A DASH-SAI pipeline is used to represent a VM NIC (ENI).
+- The VxLAN VNI is used to do the direction lookup.
+- The inner MAC address is used for pipeline lookup, a.k.a. ENI lookup (ENI).
+- Once it enters the corresponding DASH pipeline, the `outbound` pipeline will be used to process the packets sourced from the VM, while the `inbound` pipeline will be used to process the packets destined to the VM.
+
+## 5. Pipeline components
+
+### 5.1. Per-packet metadata bus
+
+First of all, since we have multiple matching stages in the pipeline, we need a way to pass the information from the matched entries in the earlier stages to the later ones to help us making final decisions on packet transformation. And this is what metadata bus is for.
```
KrisNey-MSFT, line 169:
```suggestion
First of all, since we have multiple matching stages in the pipeline, we need a way to pass the information from the matched entries in the earlier stages to the later ones to help us make final decisions on packet transformation. And this is what metadata bus is for.
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,886 @@
...
+
+At high-level, the metadata bus is a set of fields that being carried all the way through the pipeline along with the packet. It contains:
```
KrisNey-MSFT, line 171:
```suggestion
At high-level, the metadata bus is a set of fields that are carried throughout the pipeline along with the packet. It contains:
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,886 @@
...
+
+- The information from the original packet, such as encap information.
+- The information from each matched entry and related data structures, e.g., when a VNET mapping entry is matched, we will publish the information from VNET to the metadata bus.
+- The routing action and its inline parameters from the matched entry.
+
+Implementation-wise, this is similar to Packet Header Vector or Bus in NPL.
+
+### 5.2. Packet parsing
+
+In ASIC, parser is usually the first stage of the pipeline. It parses the incoming packet and extract the information into metadata bus, such as the source and destination MAC, IP, etc. And based on these information, we will invoke the rest of pipeline.
```
KrisNey-MSFT, line 181:
```suggestion
In an ASIC, the parser is usually the first stage of the pipeline. It parses the incoming packet and extracts the information into the metadata bus, such as the source and destination MAC, IP, etc. And based on this information, we will invoke the rest of pipeline.
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,886 @@
...
+
+#### 5.2.1. Multi-layer encap parsing
+
+Multi-layer encap brings the problem in the packet parsing, because of encap layer detection and overlay packet handling:
```
KrisNey-MSFT, line 185:
```suggestion
Multi-layer encap generates a problem in the packet parsing, because of the encap layer detection and overlay packet handling:
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,886 @@
...
+
+1. In ASIC, parser is always parsing from the outermost bits to innermost bits. However, in DASH, the encap stack is defined reversely: underlay1 -> underlay0 -> overlay. Because the incoming packet could have various number of encaps, we need to be able to map the encaps to the right layer.
```
KrisNey-MSFT, line 187:
```suggestion
1. In an ASIC, the parser is always parsing from the outermost bits to innermost bits. However, in DASH, the encap stack is defined reversely: underlay1 -> underlay0 -> overlay. Because the incoming packet could have various number of encaps, we need to be able to map the encaps to the right layer.
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,886 @@
...
+2. The overlay (customer) packet could be also using encaps, so we need to be able to tell which one is ours and which one is overlay.
+3. The incoming packet could have various number of encaps, and each encap doesn't really know if the inner packet is a overlay or not. For example, from the SDN pipeline basic element doc, the [Inbound from LB](./sdn-pipeline-basic-elements.md#inbound-from-lb) and [Internal Load Balancer in VNET communication](./sdn-pipeline-basic-elements.md#internal-load-balancer-in-vnet-communication) actually shares the same outer most header.
```
KrisNey-MSFT, line 189:
```suggestion
3. The incoming packet could have various number of encaps, and each encap is not aware whether the inner packet is a overlay or not. For example, from the SDN pipeline basic element doc, the [Inbound from LB](./sdn-pipeline-basic-elements.md#inbound-from-lb) and [Internal Load Balancer in VNET communication](./sdn-pipeline-basic-elements.md#internal-load-balancer-in-vnet-communication) actually shares the same outer-most header.
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,886 @@
...
+
+To solve this problems, we use 2 things for encap handling:
```
KrisNey-MSFT, line 191:
```suggestion
To solve this problems, we use 2 methods for encap handling:
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,886 @@
...
+
+1. We can specify if parser should stop parsing more encaps or not for VNI:
```
KrisNey-MSFT, line 193:
```suggestion
1. We can specify if parser should stop parsing more encaps or not for the VNI:
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,886 @@
...
+
+   ```json
+   { "DASH_SAI_VNI_TABLE|12345": { "direction": "outbound", "final_encap": true } }
+   ```
+
+   When `final_encap` is set to true, the VNI will be sent to parser to force the parser to treat the next layer as overlay packet.
+
+2. Since the parser doesn't know how many encaps beforehand, so parser can always start treating the outermost layer as underlay1. And whenever it sees an the protocol is not an encap or the VNI is unknown, it treat the packet as overlay packet. Then, after parsing is done, we will fix the encap information in metadata bus.
```
KrisNey-MSFT, line 201:
```suggestion
2. Since the parser doesn't know how many encaps existed prior, the parser can always start treating the outermost layer as underlay1. And whenever it sees that the protocol is not an encap, or the VNI is unknown, it will treat the packet as overlay packet. Then, after parsing is complete, we will fix the encap information in metadata bus.
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,886 @@
...
+
+   For example, in the [Inbound from LB](./sdn-pipeline-basic-elements.md#inbound-from-lb) case, the outer encap will start to be mapped to underlay1, the ethernet and IP part of the inner packet will be mapped into underlay0, while the TCP/UDP part will be mapped into overlay. Then, after parsing, we will extract the ethernet and IP parts in underlay0 as overlay, as well as extract the underlay1 as underlay0.
```
KrisNey-MSFT, line 203:
```suggestion
   For example, in the [Inbound from LB](./sdn-pipeline-basic-elements.md#inbound-from-lb) case, the outer encap will initially be mapped to underlay1, the ethernet and IP part of the inner packet will be mapped into underlay0, while the TCP/UDP part will be mapped into overlay. Then, after parsing, we will extract the ethernet and IP parts in underlay0 as overlay, as well as extract the underlay1 as underlay0.
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,886 @@
...
+
+### 5.3. Direction Lookup
+
+In DASH-SAI pipeline, traffic are split into 2 directions: `inbound` and `outbound`. Each direction has its own pipeline (see pipeline overview above). When a new packet arrives, we will assign a direction to the packet, then process the packet in the corresponding pipeline. This ensures us to match the flow and transform the packet in the right way.
```
KrisNey-MSFT, line 207:
```suggestion
In the DASH-SAI pipeline, traffic has 2 directions: `inbound` and `outbound`. Each direction has its own pipeline (see pipeline overview above). When a new packet arrives, we will assign a direction to the packet, then process the packet in the corresponding pipeline. This ensures we match the flow and transform the packet correctly.
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,886 @@
...
+
+### 5.4. Pipeline Lookup
+
+DASH supports multi-tenancy model for traffic handling. A single device can have multiple pipelines, and each pipeline is used to handle traffic for a specific tenant or whichever network function being modeled, such as a load balancer. When a packet arrives, besides direction lookup, we also need pipeline lookup to determine which pipeline to use for processing the packet.
```
KrisNey-MSFT, line 211:
```suggestion
DASH supports a multi-tenancy model for traffic handling. A single device can have multiple pipelines, and each pipeline is used to handle traffic for a specific tenant, or the intended network function being modeled (such as a load balancer). When a packet arrives, besides direction lookup, we also need pipeline lookup to determine which pipeline to use for processing the packet.
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,886 @@
...
+
+For example, if we like to implement a VM NIC with, then we can model one pipeline as one VM NIC, then use the inner packet MAC to find the pipeline. However, if we would like to implement a load balancer, we can use a DASH-SAI pipeline to represent a load balancer instance and use the Public IP to find the SAI pipeline.
```
KrisNey-MSFT, line 213:
```suggestion
For example, if we want to implement a VM NIC with (uncertain what this means?  a VM NIC only?), then we can model one pipeline as one VM NIC, and use the inner packet MAC to find the pipeline. However, if we would like to implement a load balancer, we can use a DASH-SAI pipeline to represent a load balancer instance and use the Public IP to find the SAI pipeline.
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,886 @@
...
+    EL --> |ENI 0| EP0
+    EL --> |ENI 1| EP1
+    EL --> |ENI 2| EP2
+
+    VL --> |LB 0| LBP0
+    VL --> |LB 1| LBP1
+    VL --> |LB 2| LBP2
+```
+
+A pipeline can also define its initial matching stages, which will be used for starting processing the packets if no flow is matched.
```
KrisNey-MSFT, line 242:
```suggestion
A pipeline can also define its initial matching stages, which will be used to start processing the packets if no flow is matched.
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,886 @@
...
+
+> NOTE:
+>
+> DASH-SAI pipeline is a logical concept. It doesn't have to be 1 to 1 mapping to a physical ASIC pipeline. The underlying implementation can be as simple as a metadata field update with a specific value, when the entry getting matched, e.g., `ENI = Inner Source/Destination MAC`.
```
KrisNey-MSFT, line 246:
```suggestion
> The DASH-SAI pipeline is a logical concept. It doesn't have to be a 1 to 1 mapping to a physical ASIC pipeline. The underlying implementation can be as simple as a metadata field update with a specific value, when the entry is matched, e.g., `ENI = Inner Source/Destination MAC`.
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,886 @@
...
+
+### 5.5. Packet Decap
+
+If a pipeline is found, before processing the packets, all outer encaps will be decap'ed, and with key information saved in metadata bus, such as encap type, source IP and VNI / GRE Key, exposing the inner most packet going through the pipeline. This simplifies the flow matching logic and also allow us to create the reverse flow properly.
```
KrisNey-MSFT, line 250: The first sentence is pretty long, is there any way to shorten it?

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,886 @@
...
+
+#### 5.5.1. Multi-layer encap handling
+
+The incoming packet to DASH pipeline could have multiple layers of encap, however, the ASIC capacity is usually limited. The more encaps we handle in parser and decaps, the more the latency will be. Furthermore, using packet recirculation is also not ideal, because it can greatly reduces the throughput and increases the latency.
```
KrisNey-MSFT, line 254:
```suggestion
The incoming packet to DASH pipeline could have multiple layers of encap, however, the ASIC capacity is usually limited. The more encaps we handle in the parser and decaps, adds latency.  Furthermore, using packet recirculation is also not ideal, because it can greatly reduce throughput and increase latency.
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,886 @@
...
+
+The limit of having 2 layers of encap at maximum is trying to help this problem and avoid the pipeline having unexpected latency.
```
KrisNey-MSFT, line 256:
```suggestion
The limit of having 2 layers of encap at maximum is an attempt to help this problem and avoid the pipeline potentially experiencing unexpected latency.
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,886 @@
...
+    "DASH_SAI_VNI_TABLE|12345": { "direction": "outbound", "stateless": false },
+    "DASH_SAI_VNI_TABLE|12346": { "direction": "outbound", "stateless": true }
+}
+```
+
+#### 5.5.3. Encap fields handling
+
+##### 5.5.3.1. Handling DSCP
+
+DASH pipeline provides 2 modes for handling the DSCP: "Preserve model" and "Pipe model".
```
KrisNey-MSFT, line 275:
```suggestion
The DASH pipeline provides 2 modes for handling the DSCP: "Preserve model" and "Pipe model".
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,886 @@
...
+- The DSCP value of overlay packet (customer packet) shall never be modified or preserved and copied to the outer header.
+
+On the other hand, pipe model works very similar to `SAI_TUNNEL_DSCP_MODE_PIPE_MODEL`:
+
+- The DSCP value of existing encaps will not be preserved.
+- When any encap being added, the DASH pipeline will specify the DSCP value, and all routing action will take this value and copy it to the encaps.
+
+If no encaps are added, say, for traffic sending to Internet, the DSCP value of overlay packet will be exposed as it is, even-if original packet arrives with a non-zero value.
+
+This gives the cloud infra full control over the DSCP value and avoid the customer packet spoofing the fields.
```
KrisNey-MSFT, line 290:
```suggestion
This gives the cloud infra full control over the DSCP value and will prevent the customer from packet spoofing the fields.
```

KrisNey-MSFT (commented):
Did a bit more...will finish later

documentation/general/dash-sai-pipeline-packet-flow.md
r12f: ouch! sorry, looks like some words are somehow lost here. i am adding them in with another commit.

documentation/general/dash-sai-pipeline-packet-flow.md
r12f: sure thing!

documentation/general/dash-sai-pipeline-packet-flow.md
r12f: i took your advice and updated this part slightly different. hope it makes sense.

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,888 @@
...
+##### 5.5.3.2. Handling TTL
+
+TTL behavior for encap shall be "pipe" model (similar to SAI_TUNNEL_TTL_MODE_PIPE_MODEL):
+
+- When adding encaps, TTL value shall be default set to 64.
+- DASH pipeline shall not modify the TTL values in the overlay packet (customer packet).
+
+#### 5.5.4. Encap preservation
+
+Sometimes, depends on the scenario to implement, the customer might want to preserve certain original encaps in the outgoing traffic. For example, say we receive a packet with structure: overlay -> underlay0 -> underlay1 -> underlay2. And we want to remove or update underlay0, preserve underlay1 and remove underlay2. This gives us the problem of handling all the CRUD combinations of all encaps, including structure changes: after removing underlay0, should underlay1 becomes underlay0 or should we keep it as underlay1? All these things affects the encap related routing actions and final packet we create.
```
KrisNey-MSFT, line 303:
```suggestion
Sometimes, depending upon the scenario to implement, the customer might want to preserve certain original encaps in the outgoing traffic. For example, say we receive a packet with structure: overlay -> underlay0 -> underlay1 -> underlay2. And we want to remove or update underlay0, preserve underlay1 and remove underlay2. This gives us the problem of handling all the CRUD combinations of all encaps, including structure changes: after removing underlay0, should underlay1 becomes underlay0 or should we keep it as underlay1? All these questions affect the encap related routing actions and final packet we create.
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,888 @@
...
+
+Since all the encap information is preserved in the metadata bus for flow creation anyway, to solve this problem, we can simply recreate them in anyway we want with routing action: `tunnel_from_encap`. It allows the source and target encap and their override value being set, which allows us to preserve the encaps in anyway we want and also ensures the clarity of final transformation that we are doing in the end.
```
KrisNey-MSFT, line 305:
```suggestion
Since all of the encap information is preserved in the metadata bus for flow creation, to solve this problem, we can simply recreate them in the manner we want using the routing action: `tunnel_from_encap`. It allows the source and target encap and their override value to be set, which allows us to preserve the encaps exactly as intended, and also ensures the clarity of final transformation performed in the end.
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,888 @@
...
+
+Please note that: Encap preservation will not affect the reverse tunnel creation as "Stateless decap vs stateful decap" section described above. A encap can be preserved as well as used in reverse tunnel creation at the same time, since they are essentially 2 different things.
+
+### 5.6. Conntrack Lookup and Update
+
+After entering a specific pipeline, the first stage will be the Conntrack Lookup stage, which does the flow lookup. If any flow is matched, the saved actions will be applied, the metering counters will be updated, and the rest of pipeline will be skipped.
```
KrisNey-MSFT, line 311:
```suggestion
After entering a specific pipeline, the first stage will be the Conntrack Lookup stage, which performs the flow lookup. If any flow is matched, the saved actions will be applied, the metering counters will be updated, and the rest of pipeline will be exited.
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,888 @@
...
+
+The core of the Conntrack Lookup and Update stage is the flow table.
+
+#### 5.6.1. Flow lookup
+
+First, let's define the flow lookup behavior.
+
+The flow lookup **MUST** use the information of inner most packet. And the matching keys can be configurated via the DASH flow APIs during the DASH initialization. This allows us to support different flow matching behaviors for certain cases. For example, in L3 level routing, we don't need to create 5 tuple flows.
```
KrisNey-MSFT, line 319:
```suggestion
The flow lookup **MUST** use the information of inner most packet. And the matching keys can be configurated via the DASH flow APIs during the DASH initialization. This allows us to support different flow matching behaviors for specific cases. For example, in L3 level routing, we don't need to create 5 tuple flows.
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,888 @@
...
+
+After the flow lookup, if a flow is matched, we will apply the saved actions in the flow direction and skip the rest of the pipeline. Otherwise, we will continue the pipeline processing.
```
KrisNey-MSFT, line 321:
```suggestion
After the flow lookup, if a flow is matched, we will apply the saved actions in the flow direction and exit the rest of the pipeline. Otherwise, we will continue the pipeline processing.
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,888 @@
...
+
+#### 5.6.2. Flow creation
+
+For new connections, after all packet transformations are applied, we will create a new flow in the flow table.
+
+Since a connection can be bi-directional, the flows needs to be created as flow pairs: the one in the same direction as the packet is usually called forwarding flow, while the other direction is called reverse flow. Of course, the flow pairs **MUST** be stored based on the pipeline direction.
```
KrisNey-MSFT, line 327:
```suggestion
Since a connection can be bi-directional, the flows need to be created as flow pairs: the one in the same direction as the packet is usually called forwarding flow, while the other direction is called reverse flow. Of course, the flow pairs **MUST** be stored based on the pipeline direction.
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,888 @@
...
+
+- When outbound pipeline creates a flow, the forwarding flow should be created on the outbound side, while reverse flow should be created on the inbound side.
```
KrisNey-MSFT, line 329:
```suggestion
- When the outbound pipeline creates a flow, the forwarding flow should be created on the outbound side, while the reverse flow should be created on the inbound side.
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,888 @@
...
+- When inbound pipeline creates a flow, it creates the flow in reversed way.
```
KrisNey-MSFT, line 330:
```suggestion
- When the inbound pipeline creates a flow, it creates the flow in reverse way.
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,888 @@
...
+
+And here is how to create the flow pairs:
+
+- The forwarding flow creation is straight forward, because all the final transformations are defined in the flow actions and metadata bus.
+- To properly create the reverse flow, we will learn the information from the original tunnels, and reverse them as return encaps, which is why we need to save all the original tunnel information when doing packet decaps.
+
+##### 5.6.2.1. Tunnel learning
+
+You might already notice that, although we uses the outer encap from the original packet to create the reverse flow, but the outer encaps are decap'ed and not used in flow matching. This creates a problem when source side fails over. Like the graph shows below, the traffic shifted from VTEP 1 (Green side) to VTEP 2 (Blue side), which changes the source IP in the encap. If we don't do anything, the existing reverse flow will continue to send the packet back to green side and causing the traffic being dropped.
```
KrisNey-MSFT, line 339:
```suggestion
You might already notice that, although we uses the outer encap from the original packet to create the reverse flow, but the outer encaps are decap'ed and not used in flow matching. This creates a problem when source side fails over. Like the graph shows below, the traffic shifted from VTEP 1 (Green side) to VTEP 2 (Blue side), which changes the source IP in the encap. If we don't do anything (such as save the information), the existing reverse flow will continue to send the packet back to green side and cause the traffic to be dropped.
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,888 @@
...
+
+![Tunnel learning](./images/dash-tunnel-learning.svg)
+
+To address this problem, the source information of each encap **MUST** be saved in the flow and compared during flow match. These information should no be part of the key, but whenever they change, we should reprocess the packet and update the saved reverse tunnel in the existing flow.
+
+##### 5.6.2.2. Asymmetrical encap handling
+
+There could be cases where the encaps are asymmetrical, which means the incoming packet and return packet uses different encaps. For example, in the case below, the encap of step 2 and 4 are not symmetrical. This means we could never recreate the flow on the DASH pipeline 2 side to tunnel the packet into the extra hop first with our current flow creation logic.
```
KrisNey-MSFT, line 347:
```suggestion
There could be cases where the encaps are asymmetrical, which means the incoming packet and return packet use different encaps. For example, in the case below, the encap of step 2 and 4 are not symmetrical. This means we could never recreate the flow on the DASH pipeline 2 side to tunnel the packet into the extra hop first with our current flow creation logic.
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,888 @@
...
+
+![Asymmetrical encap](./images/dash-asymmetrical-encap.svg)
+
+To fix this issue, a special action called `reverse_tunnel` is defined, which enables reverse side of the encaps.
+
+#### 5.6.3. Flow resimulation
+
+When certain policies are changed, we might need to update the flows that is associated with the policies. For example, when a VM is moved to another place, the VNET mapping entry will be changed, and we need to update the existing flows that is associated with the VNET mapping entry. Otherwise, the outgoing traffic will be tunneled to incorrect place. This is called flow resimulation.
```
KrisNey-MSFT, line 355:
```suggestion
In the event of a policy change, we might need to update the flows associated with the policies. For example, when a VM is moved to another location, the VNET mapping entry will be changed, and we need to update the existing flows associated with the VNET mapping entry. Otherwise, the outgoing traffic will be tunneled to an incorrect location. This is called flow resimulation.
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,888 @@
...
+
+However, the flow resimulation is not as simple as removing the flow and let the packet to going through the pipeline again. For example:
```
KrisNey-MSFT, line 357:
```suggestion
However, the flow resimulation is not as simple as removing the flow and sending the packet through the pipeline again. For example:
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,888 @@
...
+
+- When a ACL is changed in the outbound side, all flows needs to be resimulated. However, this cannot be implemented by removing all flow entries and let all flows to be recreated with the latest policy. The reason is that ACLs can be asymmetric and flows are created in pair. If the flows are removed, the inbound side traffic will be dropped immediately, because there is no inbound flow to make it bypass the ACLs.
+- Certain actions might need to bypass the flow resimulation to maintain the per flow consistency. For example, in load balancer case, we will want to ensure that flow resimulation will not causing us to forward existing connections to a different backend server.
```
KrisNey-MSFT, line 360:
```suggestion
- Certain actions might need to bypass the flow resimulation to maintain the per flow consistency. For example, in a load balancer case, we will want to ensure that flow resimulation will not cause the pipeline to forward existing connections to a different backend server.
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,888 @@
...
+
+These requires us to implement the flow resimulation in a more sophisticated way, which is not fully modeled in DASH today. But we will come back to this design later.
+
+### 5.7. Pre-pipeline ACL and Post-pipeline ACL
+
+Pre-pipeline ACL and Post-pipeline ACL are used to drop the unexpected traffic before and after the packet transformation. It works as below:
+
+1. As the high-level pipeline shows above, if an incoming packet hits a flow, it will skip matching all the ACLs. If an incoming packet is denied by a ACL, the packet will be dropped without creating a flow.
```
KrisNey-MSFT, line 368:
```suggestion
1. As the high-level pipeline shows above, if an incoming packet hits a flow (i.e. is in the FastPath), it will skip matching all the ACLs. If an incoming packet is denied by a ACL, the packet will be dropped without creating a flow.
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,888 @@
...
+2. Both outbound and inbound pipeline has its own ACL stages, and only used for matching the packets in their direction.
```
KrisNey-MSFT, line 369:
```suggestion
2. Both the outbound and inbound pipeline have their own ACL stages, and are used for matching the packets in their own direction.
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,888 @@
...
+
+In DASH-SAI pipeline, routing actions are the fundamental building blocks for packet transformations. Each routing action is designed to work as below:
+
+1. Take a specific list of metadata fields as input parameters:
+   1. For example, `staticencap` action will take the `underlay_dip`, `underlay_sip`, `encap_type` and `encap_key` to encapsulate the packet.
+   2. The parameters can come from 2 places - the metadata defined associated with the entries in each table, or the routing action definition itself. More details will be discussed in the next section.
+2. Transform the packet in a specific way, e.g., encapsulate the packet, NAT'ing the address and port, etc.
+3. Independent to other actions.
+   1. With this design, we don't have to worry about the order of the actions.
+   2. This also enables the hardware to improve the E2E pipeline latency by doing the actions in parallel.
```
KrisNey-MSFT, line 406:
```suggestion
   2. This also enables the hardware to improve the E2E pipeline latency by executing the actions in parallel.
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,888 @@
...
+
+Ideally, when DASH initializes or whenever we create a new pipeline, by simply creating multiple numbers of different types of matching stages and connecting them in different ways, we can easily implement different pipeline and network functions.
+
+| Stage 0 | Stage 1 | Stage 2 | Stage 3 | Stage 4 |
+| ------- | ------- | ------- | ------- | ------- |
+| Routing 0 | IP Mapping 0 | Port Mapping | X | X |
+| Routing 0 | Routing 1 | IP Mapping 0 | IP Mapping 1 | Port Mapping |
+| Routing 0 | IP Mapping 0 | Port Mapping | Routing 1 | IP Mapping 1 |
+
+However, in reality, it might make the pipeline hard to implement, model, debug and validate or test at this moment. For example, changing matching fields, matching type and connection dynamically on pipeline creation is simply beyond the ability of P4, which is used by our behavior model today as well as many vendor's ASIC SDKs.
```
KrisNey-MSFT, line 462: Do we want to mention P4 annotations that may help P4 bmv2?

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,888 @@
...
+
+| `action_type` value | Stage |
+| ---------------- | ----- |
+| `drop` | Drop the packet |
+| `trap` | Sending it to CPU |
+| `lpmrouting` | Skip to next routing stage |
+| `maprouting` | Skip to next mapping stage |
+| `portmaprouting` | Skip to next TCP or UDP Port Mapping stage |
+
+To help us being flexible, the transition routing actions can take a few parameters:
```
KrisNey-MSFT, line 507:
```suggestion
To help us to stay flexible, the transition routing actions can take a few parameters:
```

documentation/general/dash-sai-pipeline-packet-flow.md
```diff
@@ -0,0 +1,888 @@
...
+    {
+        "action_type": "maprouting",
+        "stage_index": 0
+    }
+]
+```
+
+##### 5.9.3.2. Stage skipping
+
+To avoid loop shows up in the pipeline, the pipeline is designed to be forward only. When transition routing action is invoked, we can simply update the target stage in the metadata, then the transition behavior can simply be described by the code below:
```
KrisNey-MSFT, line 536:
```suggestion
To avoid loops appearing in the pipeline, the pipeline is designed to be forward only. When transition routing action is invoked, we can simply update the target stage in the metadata, then the transition behavior can simply be described by the code below:
```

KrisNey-MSFT (commented):
Last of my potential edits

documentation/general/dash-sai-pipeline-packet-flow.md
r12f: p4 annotations will help generating the sai headers, but sadly, it won't be able to help much in this case. :(

KrisNey-MSFT:
[like]  Kristina Moore reacted to your message:
________________________________

marian-pritsak (approved):

budgrise (approved):

KrisNey-MSFT (approved):
approving

PR #451, opened 2023-10-21
[libsai] Add attr name logging when doing get api

kcudnik:
Using metadata add generic code to obtain attr name and log it when
attr is not supported for better debugging


chrispsommers (approved):
LGTM

kcudnik:
thanks!

PR #455, opened 2023-10-31
fix for https://github.com/sonic-net/DASH/issues/454

mgheorghe:
eni was not very descriptive and one could assume it is the eni object id, updated the name to clearly specify it is the eni mac

fixes https://github.com/sonic-net/DASH/issues/454

### DIFF
--- a/documentation/general/dash-sonic-hld.md
+++ b/documentation/general/dash-sonic-hld.md
@@ -547,7 +547,7 @@ vm_vni                   = VM VNI that is used for setting direction. Also used 
 ### 3.2.8 ROUTE LPM TABLE - OUTBOUND
 
 ``` 
-DASH_ROUTE_TABLE:{{eni}}:{{prefix}} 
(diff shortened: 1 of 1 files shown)


r12f:
looks like The dash sonic hld is moving to sonic repo while DASH repo will link to the sonic one as reference, so I wonder if we could make the change there. both prince and i should be able to help sign off: https://github.com/sonic-net/SONiC/blob/master/doc/dash/dash-sonic-hld.md. 

and here is the PR that I refer to: https://github.com/sonic-net/DASH/pull/447.

prsunny (approved):

r12f (approved):

prsunny (changes requested):
@mgheorghe , changes lgtm, but can you make the change here - https://github.com/sonic-net/SONiC/blob/master/doc/dash/dash-sonic-hld.md.

KrisNey-MSFT:
do not merge yet or cancel

mgheorghe:
canceling, was a misunderstanding

PR #460, opened 2023-11-11
Flow resimulation HLD.

r12f:
This is the follow up flow resimulation design for packet flow: https://github.com/sonic-net/DASH/pull/449.

## Motivation

In current DASH pipeline, the flow resimulation is designed to rely on a incomplete flow table (ConnTrack Lookup stage), so both ACL and mapping table update can take effect immediately. But this design could not provide a good way to:

1. Describe or leverage the fast path in the ASIC.
2. Resimulate all flows in a pipeline.
3. Provide a generic way for policy-based flow resimulation.
4. Support learning-based flow resimulation, such as tunnel learning.
5. Handle cases where we would like to maintain PCC.
6. ...

## Changes

The HLD is trying to describe how flow resimulation should work and the changes that we need in the pipeline.

The flow API doc is not merged into DASH yet. For reference, here is the PR, which will be moved into DASH repo later: <https://github.com/sonic-net/SONiC/pull/1483>.


r12f:
Hi @budgrise and @chrispsommers , I have updated the doc to add new ways for reduce the memory usage of flow tracking key in the match stage entry, as well as added the clarification for hash table. Feel free to check it out again and hopefully this addresses your concerns.

documentation/dataplane/dash-flow-resimulation.md
```diff
@@ -0,0 +1,213 @@
...
+Since the flow incarnation id is part of the flow state, it will also be synched to the standby side and updates the stored id there. Although in flow HA design, the policy will be programmed to both active and standby side, but we cannot use the id in the standby side directly, because these 2 pipelines are programmed independently, so we don't have a way to ensure that the standby one always matches the active one. To solve this, we make the standby pipeline always follows the active side, which follows the flow lifetime management design in flow HA.
+
+### 1.5. Object model change summary for full flow resimulation
+
+To summarize, the following changes are needed to implement full flow resimulation:
+
+- 2 properties needs to be added for each pipeline:
+
+    ```json
+    "DASH_SAI_ENI_TABLE|123456789012": {
```
marian-pritsak, line 66: Section 1.1 talks about pipeline level ID, while the schema defines per-ENI ID. Which one's correct? I would prefer a more specific per-ENI if it's possible, to reduce the blast radius of the ID increment.

documentation/dataplane/dash-flow-resimulation.md
```diff
@@ -0,0 +1,213 @@
...
+        "flow_incarnation_id": 0,
```
marian-pritsak, line 67: Besides ID, a max resimulation rate should be added to limit the burst of resimulated flows, as they usually compete with the CPS of new flows. A user should be able to tune this.

marian-pritsak (changes requested):

documentation/dataplane/dash-flow-resimulation.md
```diff
@@ -0,0 +1,213 @@
...
+- For simulated flows, we continue to run the later ACL and match stages instead of directly applying the actions saved in the flow.
+- In Conntrack Update stage, a new pair of flow will be generated and replaces the current flow, which finishes the process of flow resimulation.
+
+After this, the later packet will directly hit the new flow and bypass the later stages.
+
+### 1.3. Flow incarnation id overflow
+
+Some flows can be very low volume and causing flow incarnation id stops working. Say, if the flow incarnation id is 8 bits, and not it is set to 1, a low volume flow is created. And after 256 flow resimulation calls, the next packet finally arrives and see the same id, which bypasses the flow resimulation process.
+
+To solve this, whenever the incarnation id overflows, we will need to treat all flows as resimulated. Implementation-wise, this can be done by adding another bit to indicate overflow happened, then reset the flow incarnation id to 0 in each flow during the next flow aging process. After flow aging is done, we can reset the overflow bit.
```
marian-pritsak, line 47: How would the user know when to clear the overflow flag?

marian-pritsak (changes requested):

documentation/dataplane/dash-flow-resimulation.md
```diff
@@ -0,0 +1,213 @@
...
+  ```c
+  typedef enum _sai_flow_state_metadata_attr_t {
+      SAI_FLOW_ATTR_START,
+      // ...
+      SAI_FLOW_METADATA_ATTR_INCARNATION_ID, // Saved flow incarnation id.
+      SAI_FLOW_METADATA_ATTR_END
+  } sai_flow_metadata_attr_t;
+  ```
+
+## 2. Policy-based flow resimulation
```
marian-pritsak, line 84: This method makes an assumption that each data packet goes through every logical stage in the pipeline, so we can pick up stage keys along the way, which is not necessarily true. Stages may be compressed for a better performance.

For example,
a given flow used to match routing entry A, and was offloaded as such. Then the user configured an entry B with a longer prefix matching this flow. If the flow is already offloaded, there's no way of knowing that now it needs to match the entry B because routing lookup doesn't happen again for this offloaded flow.

marian-pritsak (changes requested):

documentation/dataplane/dash-flow-resimulation.md
r12f: the overflow flag will be cleaned up by data plane app after flow aging process is completed. it mostly works as full flow resimulation requested flag. maybe we should just use that instead of introducing a new flag.

ideally, we should use increase on these attributes, but SAI doesn't support this operation.

r12f: actually, rename the flag should make things more clear. will update the doc.

documentation/dataplane/dash-flow-resimulation.md
r12f: The doc is updated for this issue: 2.4. Flow resimulation with overlapping match stage key

documentation/dataplane/dash-flow-resimulation.md
r12f: the flag is renamed also explain how the workflow works.

documentation/dataplane/dash-flow-resimulation.md
r12f: updated.

documentation/dataplane/dash-flow-resimulation.md
marian-pritsak: The rate limiting takes into assumption asynchronous requests (packets are mirrored instead of queueing the original).

documentation/dataplane/dash-flow-resimulation.md
marian-pritsak: Then how does a user know the flow resimulation finished?

r12f: yea, i guess we will also need to add a callback on the switch then. otherwise, the user won't be able to know indeed.

r12f: although, what would be the granularity for reporting. and how to identify which request we are working... I need to think more on that.

documentation/dataplane/dash-flow-resimulation.md
r12f: @marian-pritsak , not sure this is nature enough for networking scenarios. although I feel token bucket is a good choice for this scenario, but it doesn't seem to be matching any existing rate limiting mechanisms.

documentation/dataplane/dash-flow-resimulation.md
```diff
@@ -0,0 +1,339 @@
...
+- Each ENI will store its own flow incarnation id, which starts from 0 and changes whenever all flows needs to resimulated by updating the SAI attribute on the ENI.
+- Each flow also stores the current id when it is created.
+
+### 1.2. Flow resimulation process
+
+With flow incarnation id, we can implement full flow resimulation as follows:
+
+- When a new packet arrives, the current flow incarnation id will be stored into the metadata bus.
+- In Conntrack Lookup stage, if we found the id stored in its flow doesn't match the one in the pipeline, we treat this flow as resimulated.
+- For simulated flows, we continue to run the later ACL and match stages instead of directly applying the actions saved in the flow.
```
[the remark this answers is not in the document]
r12f: oops. thanks! typo.

documentation/dataplane/dash-flow-resimulation.md
```diff
@@ -0,0 +1,339 @@
...
+- In Conntrack Update stage, a new pair of flow will be generated and replaces the current flow, which finishes the process of flow resimulation.
+
+After this, the later packet will directly hit the new flow and bypass the later stages.
+
+### 1.3. Flow incarnation id overflow
+
+Some flows can be very low volume and causing flow incarnation id stops working. Say, if the flow incarnation id is 8 bits, and now it is set to 1, a low volume flow is created. And after 256 flow resimulation calls, the next packet finally arrives and see the same id, which bypasses the flow resimulation process.
```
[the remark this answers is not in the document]
r12f: actually nothing particular. i was just trying to use a random non-zero number to start with, but i am horrible choosing random number, so... it ended with 1....

0 totally works! will change to that.

documentation/dataplane/dash-flow-resimulation.md
```diff
@@ -0,0 +1,339 @@
...
+   4. [2.4. Flow resimulation with overlapping match stage key](#24-flow-resimulation-with-overlapping-match-stage-key)
+   5. [2.5. Flow tracking key in flow HA](#25-flow-tracking-key-in-flow-ha)
+   6. [2.6. Object model change summary for policy-based flow resimulation](#26-object-model-change-summary-for-policy-based-flow-resimulation)
+3. [3. Learning-based flow resimulation](#3-learning-based-flow-resimulation)
+4. [4. On-demand flow resimulation](#4-on-demand-flow-resimulation)
+5. [5. Explicit per flow consistency (PCC) support](#5-explicit-per-flow-consistency-pcc-support)
+
+## 1. Full flow resimulation
+
+Full flow resimulation is the most common flow resimulation. For example, after ACL is updated, we will need to resimulate all flows for this pipeline, i.e., ENI.
```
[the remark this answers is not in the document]
r12f: updated now.

documentation/dataplane/dash-flow-resimulation.md
```diff
@@ -0,0 +1,339 @@
...
+
+- Each ENI will store its own flow incarnation id, which starts from 0 and changes whenever all flows needs to resimulated by updating the SAI attribute on the ENI.
+- Each flow also stores the current id when it is created.
+
+### 1.2. Flow resimulation process
+
+With flow incarnation id, we can implement full flow resimulation as follows:
+
+- When a new packet arrives, the current flow incarnation id will be stored into the metadata bus.
+- In Conntrack Lookup stage, if we found the id stored in its flow doesn't match the one in the pipeline, we treat this flow as resimulated.
```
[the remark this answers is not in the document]
r12f: yep! updated now.

documentation/dataplane/dash-flow-resimulation.md
```diff
@@ -0,0 +1,339 @@
...
+
+      /**
+       * @brief Flow incarnation id.
+       *
+       * @type sai_uint8_t
+       * @flags CREATE_AND_SET
+       * @default 0
+       */
+      SAI_ENI_ATTR_FLOW_INCARNATION_ID,
+
```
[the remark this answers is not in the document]
r12f: the doc is updated now.

documentation/dataplane/dash-flow-resimulation.md
```diff
@@ -0,0 +1,339 @@
...
+      /**
+       * @brief Force flow resimulation requested.
+       *
+       * @type bool
+       * @flags CREATE_AND_SET
+       * @default false
+       */
+      SAI_ENI_ATTR_FORCE_FLOW_RESIMULATION_REQUESTED,
+
```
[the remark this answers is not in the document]
r12f: the doc is updated now.

r12f:
due to new requirement showing up, i am going to close this PR for now and come back later.

PR #461, opened 2023-11-12
Add capability model for DASH pipeline and matching stages.

r12f:
This change is a follow up update for the new packet flow.

Although, the logical pipeline or behavior model allows us to set any metadata in any stage or doing any routing actions that we want, the ASIC can have certain limitations, either the functionality or the resource. So, we should allow technology providers to tell the user what is supported via capability.


KrisNey-MSFT:
@r12f - do we need to cover this in the next meeting?  Or is it ready for a Merge?

PR #470, opened 2023-12-07
Improve readability of sai_api_gen.py for SAI header file generation.

r12f:
This PR will not change any generated SAI header files but only update the sai_api_gen.py file for better readability.

After this change, all header files are lib files are output with the exactly the same format:

SAI folder:
image

Lib:
image


marian-pritsak (approved):

KrisNey-MSFT:
Reorganized sections to more easily see the hierarchy and become more obvious.  Much easier changes in the future; for this one no functionality will change, all produced files will remain unchanged.

PR #479, opened 2023-12-12
[sai-gen] Deprecate the name annotations in favor of SaiTable

r12f:
This PR has 2 updates:

1. Deprecate the `@name` annotations in favor of `@SaiTable`.
2. Add `api_order` attribute in `@SaiTable` annotation to help us enforce the order of the generated APIs.

The second change is a must have, otherwise for any API set that contains multiple type of entries (generated from multiple tables), the order could be changed between changes and break the ABI compatibility as the screenshot shows below:

Snipaste_2023-12-12_12-24-41

The doc is also updated to capture the use of this new attribute.

For generated content updates:

1. No updates on SAI header files:

    ```bash
    r12f@r12f-dl380:~/data/code/sonic/DASH/dash-pipeline
    $ diff SAI/SAI/experimental/ ~/data/code/sonic/DASH-exp/dash-pipeline/SAI/SAI/experimental/
    r12f@r12f-dl380:~/data/code/sonic/DASH/dash-pipeline
    ```

2. For libsai, table id will be updated as previous change. 

    ```bash
    r12f@r12f-dl380:~/data/code/sonic/DASH/dash-pipeline
    $ diff SAI/SAI/experimental/ ~/data/code/sonic/DASH-exp/dash-pipeline/SAI/SAI/experimental/
    r12f@r12f-dl380:~/data/code/sonic/DASH/dash-pipeline
    $ diff SAI/lib/ ~/data/code/sonic/DASH-exp/dash-pipeline/SAI/lib/
    diff SAI/lib/saidashacl.cpp /home/r12f/data/code/sonic/DASH-exp/dash-pipeline/SAI/lib/saidashacl.cpp
    31c31
    <     tableId = 50200087;
    ---
    >     tableId = 45323240;
    diff SAI/lib/saidashdirectionlookup.cpp /home/r12f/data/code/sonic/DASH-exp/dash-pipeline/SAI/lib/saidashdirectionlookup.cpp
    19c19
    <     pi_p4_id_t tableId = 45670434;
    ---
    >     pi_p4_id_t tableId = 38960243;
    110c110
    <     pi_p4_id_t tableId = 45670434;
    ---
    >     pi_p4_id_t tableId = 38960243;
    diff SAI/lib/saidasheni.cpp /home/r12f/data/code/sonic/DASH-exp/dash-pipeline/SAI/lib/saidasheni.cpp
    19c19
    <     pi_p4_id_t tableId = 46804748;
    ---
    >     pi_p4_id_t tableId = 38612462;
[...]

PR #480, opened 2023-12-13
[sai-gen] Deprecate name attribute for all match keys and action parameters, remove type guessing heuristics on object parent names.

r12f:
This change is based on #479 , hence including all changes in the PR.

No update on SAI headers after this change:

```bash
```

2 comments are updated in lib file, but looks like the previous result is wrong:
image

Take route_vnet as an example, it indeed only take 4 parameters, as ip_is_v6 field doesn't count:
image

### DIFF
--- a/dash-pipeline/SAI/sai_api_gen.py
+++ b/dash-pipeline/SAI/sai_api_gen.py
@@ -158,56 +158,51 @@ class SAITypeSolver:
         if match_type == 'exact' or match_type == 'optional' or match_type == 'ternary':
             return  SAITypeSolver.get_object_sai_type(key_size, key_parent_name, key_name)
         elif match_type == 'lpm':
-            return SAITypeSolver.__get_lpm_match_key_sai_type(key_size, key_parent_name, key_name)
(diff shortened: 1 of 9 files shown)


r12f:
Please ignore this PR for review until #479 is checked in.

r12f:
looks like in certain cases, `@name` is required:

image

chrispsommers:
> looks like in certain cases, `@name` is required: (@chrispsommers as FYI)

That's fine, there's nothing wrong with the @name annotation, it is widely used and guides P4 artifact generation. The only objection in the past was using the pipe symbol to signify SAI code generation; annotations are preferred.

chrispsommers (approved):
LGTM, thanks for this contribution.

r12f:
Rebased on top of #479 without any code changes.

r12f:
> > looks like in certain cases, `@name` is required: (@chrispsommers as FYI)
> 
> That's fine, there's nothing wrong with the @name annotation, it is widely used and guides P4 artifact generation. The only objection in the past was using the pipe symbol to signify SAI code generation; annotations are preferred.

Yep, makes sense!

PR #485, opened 2023-12-13
[sai-gen] Add range match type support in SAI generation script.

r12f:
The range match is frequently used for cases like port ranges and etc.

This change doesn't change anything in the pipeline, but only add the functionality support for future cases.

More can be found in:

- DASH packet flow: https://github.com/sonic-net/DASH/blob/main/documentation/general/dash-sai-pipeline-packet-flow.md
- Usage in SAI ACL (not DASH): https://github.com/opencomputeproject/SAI/blob/master/inc/saiacl.h
- SAI types: https://github.com/opencomputeproject/SAI/blob/master/inc/saitypes.h

### DIFF
--- a/dash-pipeline/SAI/sai_api_gen.py
+++ b/dash-pipeline/SAI/sai_api_gen.py
@@ -104,6 +104,7 @@ class SAITypeSolver:
         "sai_u16_list_t": SAITypeInfo("sai_u16_list_t", "u16list"),
         "sai_u32_list_t": SAITypeInfo("sai_u32_list_t", "u32list"),
         "sai_ip_prefix_list_t": SAITypeInfo("sai_ip_prefix_list_t", "ipprefixlist"),
+        "sai_u32_range_t": SAITypeInfo("sai_u32_range_t", "u32range"),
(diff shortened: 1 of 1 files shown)


chrispsommers (commented):
LGTM but I notice that CI did not get triggered? I would think the rules in https://github.com/sonic-net/DASH/blob/main/.github/workflows/dash-bmv2-ci.yml and the (dpdk version as well) should trigger a CI pipeline, in particular: 
```
      - 'dash-pipeline/**'
```
Without a CI run, one can't tell from the PR if there are any regressions,

r12f:
yea, this is interesting... Let me check the CI and do another push.

r12f:
looks like a UI problem. the CI actually ran, but didn't get picked up:

image

chrispsommers (approved):
LGTM

r12f:
there you go. made a very minor update in the same file, and CI is working now.

r12f:
thanks Chris!

r12f:
all CI are passed (officially) now. :D

PR #486, opened 2023-12-14
[sai-gen] Update ACL tables type and match type annotations for SAI header generation

r12f:
This PR changes 3 things:

1. Update the type annotation of ACL table match keys, so we can generate the right types for ACLs.
2. Add `match_type` annotation for match keys, so we can fix the `optional` match kind that we use to make P4 compiler happy.
3. Update SAI header generation script to handle the new annotation.

The end results is shown as below. The comment and type are now updated to the same as opencompute SAI:

image
(no changes showing up in git diff)

To compare with previous generated content:

- Header files:
    
    ```bash
    diff SAI/SAI/experimental/saiexperimentaldashacl.h /home/r12f/data/code/sonic/DASH-exp/dash-pipeline/SAI/SAI/experimental/saiexperimentaldashacl.h
    114c114
    <      * @brief List matched key dip
    ---
    >      * @brief Optional matched key dip
    ...
    ```
    
    image

- libsai (creating adapter that converts list / range list to optional for bmv2):

    image

### DIFF
--- a/dash-pipeline/SAI/sai_api_gen.py
+++ b/dash-pipeline/SAI/sai_api_gen.py
@@ -155,9 +155,9 @@ class SAITypeSolver:
         elif match_type == 'list':
             return SAITypeSolver.__get_list_match_key_sai_type(key_size)
         elif match_type == 'range':
-            return SAITypeSolver.__get_range_sai_type(key_size)
+            return SAITypeSolver.__get_range_match_key_sai_type(key_size)
         elif match_type == 'range_list':
-            return SAITypeSolver.__get_range_list_sai_type(key_size)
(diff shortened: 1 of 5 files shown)


r12f:
PR is not ready for review. Need to discuss what is the right way for libsai changes. My feeling is to convert the input to optional, because this is what BMv2 supports.

r12f:
discussed in today's DASH BM meeting today. For BMv2, we will convert the ACL input to optional by taking the first item of the list.

r12f:
OK, the tests are fixed now. this PR is ready for review. @marian-pritsak @kcudnik

chrispsommers (approved):

kcudnik (approved):

PR #489, opened 2023-12-21
[sai-gen] Add type hints for SAI generation scripts.

r12f:
Add type hints to make the code more maintainable since it will be even less likely to go wrong.

As previous refactoring changes, no updates on generated files:

image

### DIFF
--- a/dash-pipeline/SAI/sai_api_gen.py
+++ b/dash-pipeline/SAI/sai_api_gen.py
@@ -40,7 +40,7 @@ SAI_API_ORDER_TAG: str = 'api_order'
 #
 # SAI parser decorators:
 #
-def sai_parser_from_p4rt(cls: 'SAIObject'):
+def sai_parser_from_p4rt(cls: type['SAIObject']):
     @staticmethod
     def create(p4rt_value, *args, **kwargs):
         sai_object = cls()
--- a/dash-pipeline/SAI/sai_api_gen.py
+++ b/dash-pipeline/SAI/sai_api_gen.py
@@ -6,40 +6,41 @@ try:
     import argparse
     import copy
     from jinja2 import Template, Environment, FileSystemLoader
+    from typing import (Any, Dict, List, Optional)
 except ImportError as ie:
     print("Import failed for " + ie.name)
     exit(1)
 
-NAME_TAG = 'name'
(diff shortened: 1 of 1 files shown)


kcudnik:
Please fix build errors

r12f:
ok @kcudnik , this change is now ready for review.

chrispsommers (approved):
The code has changed so much over time to the extent it's a new program, and I don't claim to understand it entirely anymore. But I like the fact it's employing more modern practices like typing. Ultimately, it should require little maintenance and the main effort going forward should consist of writing properly-annotated P4 code, with an occassional update to the codegen to handle a new requirement or annotation.

Ironically, I can see some potential for adding heuristics to the codegen to detect missing or improper annotations, but in a way that's going backwards to making it "understand" dash code!

Well-done, thanks for the massive overhaul.

r12f:
thanks Chris!

> but in a way that's going backwards to making it "understand" dash code!

and yep, this is something I am working on right now for counters. Basically, to generate the counter attributes, we have to add unused parameters into table action parameters, which is really weird to understand. a more proper way is actually to parse the p4 bmv2 program and find the counter writes to build the relationship.

i am working on 2 PRs, one add SAICounter annotation to make the counter in use. another one to parse the P4 program and build the relationships. this will makes more sense than the heuristics we had, such as the naming patterns, or the number of fields that an action has and etc.

r12f:
adding one example as reference:

```
#ifdef TARGET_BMV2_V1MODEL
    counter(MAX_METER_BUCKETS, CounterType.bytes) meter_bucket_inbound;
    counter(MAX_METER_BUCKETS, CounterType.bytes) meter_bucket_outbound;
#endif // TARGET_BMV2_V1MODEL
    action meter_bucket_action(
            @SaiVal[type="sai_uint64_t", isreadonly="true"] bit<64> outbound_bytes_counter,
            @SaiVal[type="sai_uint64_t", isreadonly="true"] bit<64> inbound_bytes_counter,
            @SaiVal[type="sai_uint32_t", skipattr="true"] bit<32> meter_bucket_index) {
        // read only counters for SAI api generation only
        meta.meter_bucket_index = meter_bucket_index;
    }
```

PR #497, opened 2024-01-04
[sai-gen] Add order annotation for attributes to support ordering SAI attributes for ABI compatibility

r12f:
## Problem 

Currently, DASH SAI attributes are generated based on the order of table keys and action parameters. When multiple types of objects are mixed together, adding a new attributes could cause new SAI attributes being inserted in the middle of the attribute list and breaks the ABI compatibility.

This change introduces 2 things:

- adds `order` attribute in `@SaiVal` annotation, so we could sort the attributes in the way that keeps the ABI compatibility.
- renames `api_order` to `order` in `@SaiTable` annotation, to keep the naming the same way.

The current change will not change the generated content today, as screenshot shows below:
image

### DIFF
--- a/dash-pipeline/SAI/sai_api_gen.py
+++ b/dash-pipeline/SAI/sai_api_gen.py
@@ -6,7 +6,7 @@ try:
     import argparse
     import copy
     from jinja2 import Template, Environment, FileSystemLoader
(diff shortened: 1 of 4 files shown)


marian-pritsak (approved):

r12f:
thanks Marian!

PR #499, opened 2024-01-05
[sai-gen] Make SAI generator understand P4 IR to fetch counter references / callers.

r12f:
## Problem

Currently, the SAI attributes will be generated from table keys, action parameters and counters. Table keys and action parameters are naturally connected to the tables directly or indirectly in the P4 runtime json. However, counters are not, so we don't have enough information to find which SAI API we should add the attribute.

To solve this, we have already added an annotation in SaiCounter, to specify the action names, so we can correlate the counters and its tables. However, this is tedious to do, if we have a lot of counters to add.

This change parses the P4 IR json file to find which action this counter has been called to, so we can associate the counters and its action and tables automatically. 

Take an example, say we add a new counter in the metering action to calculate total bytes:

image

And without specifying the action names, like the other counters, the SAI attribute will be generated in the right place, because we now understand the program.

image

We choose the P4 IR as the input instead of the bmv2 json, because this approach is backend independent, and can apply to other backends too, such as dpdk, p4tc and etc, just like the p4 runtime json. So, if we decide to change our BM from BMv2 to other implementation, this program will still work.

This change will help us add more counter for future scenarios, such as HA.

### DIFF
--- a/dash-pipeline/SAI/Makefile
+++ b/dash-pipeline/SAI/Makefile
@@ -2,6 +2,7 @@
 all: copysrc
 	./sai_api_gen.py \
 		/bmv2/dash_pipeline.bmv2/dash_pipeline_p4rt.json \
+		--ir /bmv2/dash_pipeline.bmv2/dash_pipeline_ir.json \
 		--ignore-tables=appliance,eni_meter,slb_decap \
 		dash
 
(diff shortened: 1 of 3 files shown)


r12f:
@chrispsommers , as we chatted before, now we are smart enough to understand the P4 program : D

marian-pritsak (approved):

r12f:
thanks a lot Marian!

PR #501, opened 2024-01-09
[sai-gen] Add support to packets and packet_and_bytes counters, also make counter generation to be adaptive to its type.

r12f:
## Problem

Currently, SAI API generation is not leveraging the counter type information, but blindly use whatever specified in the `name` annotation. This leads to 2 problems:

1. We could accidentally generate incorrect SAI headers, e.g. name says it is a byte counter, but in fact it is a packet counter in our BM. 
2. We could not support counters that tracks multiple things, i.e. packets and bytes.

This change uses the type info in the counter to generate the correct SAI attributes, also fork the counter when the single counter in BM tracks multiple things.

As the screenshot shows below, after the change, if we change the counter type of the meter_bucket_outbound, a new SAI attribute will be created for it.

image

This change will not change any of existing SAI header files that we generated, but can be used for future scenarios.

### DIFF
--- a/dash-pipeline/SAI/sai_api_gen.py
+++ b/dash-pipeline/SAI/sai_api_gen.py
@@ -1,5 +1,8 @@
 #!/usr/bin/env python3
 
+from typing import Iterator
+
+
 try:
     import os
     import json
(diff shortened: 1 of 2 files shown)


chrispsommers (approved):
LGTM

r12f:
Woot! Thanks @chrispsommers ! Also - happy new year!

KrisNey-MSFT:
Closing as items is Merged

PR #503, opened 2024-01-10
[sai-gen] Add counter id generation support for counters.

r12f:
## Problem

Currently, DASH only supports generating counters as SAI attributes upon DASH APIs, such as metering buckets. This might be ok for new things that has dedicated DASH concepts, i.e., metering buckets. However, for other more generic counters used for collecting telemetry, this approach could leads to feature missing and not fully utilizing what we already have in SAI and SONiC. For example, no get-and-reset operation support and need to build dedicated pipeline for pulling the counters and feeds them into the telemetry.

This change adds support to generate the counters as [SAI generic counters](https://github.com/opencomputeproject/SAI/blob/master/doc/SAI-Proposal-Generic-Counters.md). Instead of generate the counters as readonly attributes, we generate a counter id for it. Then, we can leverage all the infrastructure already built for counters.

As the screenshot shows below, after removing `as_attr="true"` for metering_bucket_outbound, a counter id will be generated instead of an readonly counter attribute. It also supports handling the packets-and-bytes counter too, because generic counter already collects both information.

image

### DIFF
--- a/dash-pipeline/SAI/sai_api_gen.py
+++ b/dash-pipeline/SAI/sai_api_gen.py
@@ -546,7 +546,7 @@ class SAICounter(SAIAPITableAttribute):
         if self.as_attr:
             counter_storage_type = SAITypeSolver.get_object_sai_type(self.bitwidth)
 
-        # Otherwise, this counter should be linked to a SAI counter using a object it.
+        # Otherwise, this counter should be linked to a SAI counter using an object ID.
         # In this case, the type needs to be sai_object_id_t.
         else:
             counter_storage_type = SAITypeSolver.get_sai_type("sai_object_id_t")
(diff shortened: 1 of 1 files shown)


dash-pipeline/SAI/sai_api_gen.py
```diff
@@ -536,9 +541,20 @@ def parse_p4rt(self, p4rt_counter: Dict[str, Any], var_ref_graph: P4VarRefGraph)
...
 
-        counter_storage_type = SAITypeSolver.get_object_sai_type(self.bitwidth)
-        self.type = counter_storage_type.name
-        self.field = counter_storage_type.sai_attribute_value_field
+        # If this counter is marked as SAI attributes, then we need to generate dedicated SAI attributes for this counter.
+        # In this case, the type needs to be created based on the size.
+        if self.as_attr:
+            counter_storage_type = SAITypeSolver.get_object_sai_type(self.bitwidth)
+
+        # Otherwise, this counter should be linked to a SAI counter using a object it.
```
chrispsommers, line 549: object id -> object ID ?

r12f: thanks for the review! updated now.

chrispsommers:
Perhaps you explain this is in the WG mtg, you've made a number of enhancements recently and I think it'd be good to review for the community. Thx.

r12f:
yea! that sounds great to me! i can do a summary on how the latest SAI generator works and all the enhancements I did recently (so far, I still won't call it new, since the basic logic is still the same :D)

not sure if tomorrow is a good time, if not, next next week should be good, if we have time.

chrispsommers (approved):
lgtm

r12f:
Woot! Thank you so much, Chris!

PR #505, opened 2024-01-18
[sai] Update SAI for adding port extension support.

r12f:
Updating SAI commit to current latest master: 7bffa33c7ad27960fe8c6baabf3605a2edd5f2d4.

This is for adding the port extension support, which includes both port attributes and port stats. The port extensions can be used for adding data path counters on port level, so we can have a higher layer of view.

See commit <https://github.com/opencomputeproject/SAI/commit/7bffa33c7ad27960fe8c6baabf3605a2edd5f2d4> for more details.


kcudnik (approved):

chrispsommers (approved):
LGTM

r12f:
thanks a lot Kamil and Chris!

PR #507, opened 2024-01-19
[Draft] Add flow API

zhixiongniu:
This PR introduces the abstraction of the DASH Flow API, encompassing both the flow table and flow entry, within the bmv2 environment.

For a preview of the SAI introduced by this PR, please refer to the following branch:
 [SAI Flow API Preview Branch](https://github.com/zhixiongniu/SAI/tree/zhixiongniu/flow_api_preview).

### DIFF
--- a/dash-pipeline/dockerfiles/DOCKER_SAI_CHALLENGER_CLIENT_BLDR_IMG.env
+++ b/dash-pipeline/dockerfiles/DOCKER_SAI_CHALLENGER_CLIENT_BLDR_IMG.env
@@ -2,4 +2,4 @@
 # Changing this will cause build/publish to occur in CI actions
 export DASH_ACR_REGISTRY=sonicdash.azurecr.io
 export DOCKER_SAI_CHALLENGER_CLIENT_BLDR_IMG_NAME=${DASH_ACR_REGISTRY}/dash-saichallenger-client-bldr
-export DOCKER_SAI_CHALLENGER_CLIENT_BLDR_IMG_CTAG?=230601
+export DOCKER_SAI_CHALLENGER_CLIENT_BLDR_IMG_CTAG?=240216
--- a/dash-pipeline/dockerfiles/Dockerfile.saichallenger-client
+++ b/dash-pipeline/dockerfiles/Dockerfile.saichallenger-client
@@ -1,4 +1,4 @@
-FROM sonicdash.azurecr.io/dash-saichallenger-client-bldr:230601
+FROM sonicdash.azurecr.io/dash-saichallenger-client-bldr:240216
 
 ENV SAI_CHALLENGER_PATH /sai-challenger
 ENV DASH_PATH /dash
--- a/dash-pipeline/dockerfiles/Dockerfile.saichallenger-client-bldr
+++ b/dash-pipeline/dockerfiles/Dockerfile.saichallenger-client-bldr
@@ -1,6 +1,6 @@
 # Requires <url of sai-challenger branch> <commit sha> or something
-# sc-client:230607
-FROM sc-client
+# sc-client:240216
+FROM sc-client:buster
 
 ADD tests/ /tests/
 
@@ -9,6 +9,10 @@ ADD SAI/SAI/test/ptf /SAI/test/ptf
 
 # Install PTF test framework & test-cases from SAI repo
 ADD SAI/SAI/ptf /SAI/ptf/
+
+# Install dependencies for dpugen==0.0.5
+RUN apt-get update && \
+    apt-get install -y python3-lxml
 
 # Install the python libraries
 RUN python3 -m pip install -r /tests/requirements.txt && \


dash-pipeline/SAI/templates/saiapi.h.j2
```diff
@@ -381,6 +381,33 @@ typedef sai_status_t (*sai_bulk_remove_{{ table.name }}_fn)(
...
+ * @param[in] mode Bulk operation error handling mode.
+ * @param[out] object_statuses List of status for every object. Caller needs to
+ * allocate the buffer
+ *
+ * @return #SAI_STATUS_SUCCESS on success when all objects are created or
+ * #SAI_STATUS_FAILURE when any of the objects fails to create. When there is
+ * failure, Caller is expected to go through the list of returned statuses to
+ * find out which fails and which succeeds.
+ */
+typedef sai_status_t (*sai_bulk_query_{{ table.name }}_fn)(
```
r12f, line 403: rename bulk_query to bulk_get, otherwise it could not pass the SAI checkers.

dash-pipeline/SAI/templates/saiapi.h.j2
```diff
@@ -381,6 +381,33 @@ typedef sai_status_t (*sai_bulk_remove_{{ table.name }}_fn)(
...
+/**
+ * @brief Bulk query {{ sai_api.app_name }}_{{ table.name }}
+ *
+ * @param[in] object_count Number of objects to create
+ * @param[in] {{ table.name }} List of object to create
+ * @param[in] attr_count List of attr_count. Caller passes the number
+ *    of attribute for each object to create.
+ * @param[in] attr_list List of attributes for every object.
+ * @param[in] mode Bulk operation error handling mode.
+ * @param[out] object_statuses List of status for every object. Caller needs to
```
r12f, line 395: since bulk get API is essentially a filter-based query, hence "object_statuses" is not intuitive to add here, so do you mind to add some comment here on how we will use it (in which case, we will get failure). it will be very helpful.

dash-pipeline/bmv2/dash_flow.p4
```diff
@@ -0,0 +1,60 @@
...
+        bit<32> flow_dest_pa,
+        bit<32> flow_metering_class,
+        bit<32> flow_rewrite_info,
+        bit<32> flow_vendor_metadata) {
+    }
+
+    @SaiTable[name = "flow_table", api = "dash_flow", api_order = 0, isobject="true"]
+    table flow_table {
+        key = {
+            meta.eni_id : exact @SaiVal[type = "sai_object_id_t"];
```
r12f, line 32: I think I found a problem if we put eni_id as an attribute to the flow table directly - it will be hard to support DPU-level bulk sync.

The reason is that with current approach, the flow table is bound to the ENI level - each ENI will have its own flow table. Hence, in order to implement DPU-level bulk sync, we will need to call bulk get API on all flow table - essentially use ENI-level bulk sync to implement DPU-level bulk sync. This might not be supported.

A better way to implement this would be:
1. Create flow table without ENI, but it can have all the attributes to set which field to enable in the flow keys, and ENI is one of them.
2. Add flow table id on the ENI, so we can bind the flow table on all ENIs (DPU-level) or bind the flow table to a single ENI (ENI-level). In the future, we can even bind multiple flow table to a ENI, which is very flexiable.

With this, to implement DPU-level bulk sync, all we need is to bulk get all flow entries on the DPU-level flow table, and call it done.

dash-pipeline/SAI/templates/saiapi.h.j2
```diff
@@ -381,6 +381,33 @@ typedef sai_status_t (*sai_bulk_remove_{{ table.name }}_fn)(
...
+ * failure, Caller is expected to go through the list of returned statuses to
+ * find out which fails and which succeeds.
+ */
+typedef sai_status_t (*sai_bulk_query_{{ table.name }}_fn)(
+        _In_ uint32_t object_count,
+        _In_ const sai_{{ table.name }}_t *{{ table.name }},
+        _In_ const uint32_t *attr_count,
+        _In_ const sai_attribute_t **attr_list,
+        _In_ sai_bulk_op_error_mode_t mode,
+        _Out_ sai_status_t *object_statuses);
```
r12f, line 409: I believe we still need the inout attributes for handling the direct query case.

dash-pipeline/SAI/sai_api_gen.py
```diff
@@ -769,6 +769,7 @@ def __init__(self):
         self.stage: Optional[str] = None
         self.is_object: Optional[str] = None
         self.api_type: Optional[str] = None
+        self.enable_query: Optional[str] = None
```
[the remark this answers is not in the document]
zhixiongniu: > enable_bulk_get_api

Done, thanks.

zhixiongniu:
So sorry. I closed this PR accidentally. 
This new PR is: https://github.com/sonic-net/DASH/pull/524

PR #508, opened 2024-01-26
[sai-gen] Add counter id generation support on port level

r12f:
## Problem

Currently, whenever a counter is used in a control block and not associated to any tables, we won't be able to handle the SAI header generation for it, although we know this counter will be mostly used for tracking the DPU level counts.

This change adds counter id generation support on port level. Whenever the above case happens, a new counter id attribute will be generated on port level.

Take an example - if we add a counter to track total bytes in DASH as below:
image
image

The following counter id attribute will be generated:
image

### DIFF
--- a/dash-pipeline/SAI/sai_api_gen.py
+++ b/dash-pipeline/SAI/sai_api_gen.py
@@ -1232,7 +1232,6 @@ class SAIGenerator:
                     if sai_counter_port_attr_name not in all_port_attrs:
                         new_port_counters.append(sai_counter)
 
-        print("Is first port attr: " + str(is_first_attr) + ", all port attrs: " + str(all_port_attrs))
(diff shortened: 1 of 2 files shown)


kcudnik (approved):

chrispsommers (approved):
LGTM, do you intend to utilize this in some p4 code soon?

r12f:
thanks Kamil and Chris!

> LGTM, do you intend to utilize this in some p4 code soon?

yea, this is made to support the port level counter id in the fast path design doc as well as counters in HA. and to avoid the final PR being too large, I split the change into these prerequisite PRs.

I have another one with stats generation, and after that i will send the fast path SAI API one out. Hopefully this can help connects the dots with the recent PRs without any model updates.

PR #509, opened 2024-02-06
Update HLD for PL scenario

theasianpianist:
- Add `routing_appliance_id` field to VNET mapping table
- Update PL encoding format
- Remove 'key' attribute in routing type table in favor of 'vni'
- Fix SIP encodings in PL example


prsunny (approved):

r12f:
The routing appliance id might need to be updated later, so do we wait for that change to be finalized or we go first and change later?

r12f (approved):

r12f:
> The routing appliance id might need to be updated later, so do we wait for that change to be finalized or we go first and change later? @prsunny

actually, np. this is more like a fix on the current HLD to make things aligned, not really change any behavior or design. lgtm! thanks @theasianpianist !

PR #512, opened 2024-02-09
Changes to support most recent version of p4c-dpdk

jafingerhut:


jafingerhut:
Note: With the version of the p4c-dpdk binary being used in the DASH repo as of 2024-Feb-09, that command will fail with the next `--context` command line option that I have added in the Makefile.

These changes should compile without errors from `p4c-dpdk` if that executable is updated to a version built from a more recent version of the p4c-dpdk source code, e.g. some time in 2024, but latest should be good.

chrispsommers (approved):
LGTM thanks!

PR #514, opened 2024-02-11
[sai-gen] Add support for generating SAI stats APIs for counters.

r12f:
## Problem

DASH has scaling requirement that requires technology providers to support huge number of objects, e.g. 8M of CA-PA mappings. If we need to add counters for these objects by using generic SAI counters, we will generate large number of counter ids, which might not be ideal.

Also some counters we are collecting are not data path counters, but more like hit counts, which is not what generic SAI counter is modeling.

For these cases, we can support generating the counters as SAI stats and leverage existing SAI stats API design for querying the counters back.

This change updated sai_api_gen.py with 2 changes:
1. Rename `as_attr` to `attr_type` for supporting more types of counter generation.
2. Updated the counter generation logic and its templates for generating SAI stats APIs.

This feature will be used in the incoming fast path API generations.

Take metering bucket as an example to illustrate this change, say we updated its counters to capture both packet and bytes, change its type to stats, then add another counter for tracking all traffic, not on metering bucket level:
image

This will results in the both metering bucket counter being generated as stats enum:
image

Once an table as any stats enum, we will generate the stats APIs:
image
image

On port level, it will generate the overall counters:
image

However, we are not generating a working libsai at this moment, which is the same as the counters today. This work will be postponed to later time.
image

image

### DIFF
--- a/dash-pipeline/SAI/sai_api_gen.py
+++ b/dash-pipeline/SAI/sai_api_gen.py
@@ -767,6 +767,7 @@ class SAIAPITableData(SAIObject):
         self.counters: List[SAICounter] = []
         self.with_counters: str = 'false'
         self.sai_attributes: List[SAIAPITableAttribute] = []
+        self.sai_stats: List[SAIAPITableAttribute] = []
 
         # Extra properties from annotations
         self.stage: Optional[str] = None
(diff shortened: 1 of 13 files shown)


r12f:
Here is the preview of the fast path SAI API change that built on top of this feature (for header generation only):

- DASH p4: https://github.com/r12f/DASH/commit/ee0a078b1a2c504517763bd7905b540678cd1cb9
- SAI API: https://github.com/r12f/SAI/commit/d0e29049664945a78e3569193d9280fde04654a3

marian-pritsak (approved):

r12f:
thanks for the review, Marian! will get this merged and sharing the fast path SAI update out.

PR #515, opened 2024-02-13
Add PL NSG

marian-pritsak:


KrisNey-MSFT:
This is about adding a 2nd layer of encapsulation - see links

marian-pritsak:
SAI Attributes for the tunnel object:
```
/**
 * @brief Attribute ID for dash_tunnel_tunnel1
 */
typedef enum _sai_tunnel1_attr_t
{
    /**
     * @brief Start of attributes
     */
    SAI_TUNNEL1_ATTR_START,

    /**
     * @brief Action set_tunnel1_attrs parameter UNDERLAY_DIP
     *
     * @type sai_ip_address_t
     * @flags CREATE_AND_SET
     * @default 0.0.0.0
     */
    SAI_TUNNEL1_ATTR_UNDERLAY_DIP = SAI_TUNNEL1_ATTR_START,

    /**
     * @brief Action set_tunnel1_attrs parameter DASH_ENCAPSULATION
     *
     * @type sai_dash_encapsulation_t
     * @flags CREATE_AND_SET
     * @default SAI_DASH_ENCAPSULATION_VXLAN
     */
    SAI_TUNNEL1_ATTR_DASH_ENCAPSULATION,

    /**
     * @brief Action set_tunnel1_attrs parameter TUNNEL_KEY
     *
     * @type sai_uint32_t
     * @flags CREATE_AND_SET
     * @default 0
     */
    SAI_TUNNEL1_ATTR_TUNNEL_KEY,

    /**
     * @brief End of attributes
     */
    SAI_TUNNEL1_ATTR_END,

    /** Custom range base value */
    SAI_TUNNEL1_ATTR_CUSTOM_RANGE_START = 0x10000000,

    /** End of custom range base */
    SAI_TUNNEL1_ATTR_CUSTOM_RANGE_END,

} sai_tunnel1_attr_t;
```

KrisNey-MSFT:
@r12f @ashutosh-agrawal   agreed to review/merge

dash-pipeline/bmv2/stages/tunnel1.p4
```diff
@@ -0,0 +1,41 @@
...
+    inout headers_t hdr,
+    inout metadata_t meta)
+{
+    action set_tunnel1_attrs(
+            @SaiVal[type="sai_ip_address_t"]
+            IPv4Address underlay_dip,
+            @SaiVal[type="sai_dash_encapsulation_t", default_value="SAI_DASH_ENCAPSULATION_VXLAN"]
+            dash_encapsulation_t dash_encapsulation,
+            bit<24> tunnel_key) {
+    push_action_static_encap(hdr = hdr,
```
ashutosh-agrawal, line 14: Would this overwrite the push_action_static_encap action performed in set_private_link_mapping action?

dash-pipeline/bmv2/stages/tunnel1.p4
```diff
@@ -0,0 +1,41 @@
...
+            meta.tunnel1_id : exact @SaiVal[type="sai_object_id_t"];
+        }
+
+        actions = {
+            set_tunnel1_attrs;
+        }
+    }
+
+    apply {
+        if (tunnel1.apply().hit) {
```
ashutosh-agrawal, line 35: This do_action_static_encap will be called before the nat action in routing_action_apply block. So, we will end up performing vxlan_encap followed by nat4to6 and then NVGRE encap. But the correct order would be nat4to6 followed by nvgre encap and then vxlan encap. Am I not reading the code correctly?

dash-pipeline/bmv2/dash_routing_types.p4
```diff
@@ -193,4 +193,30 @@ action set_private_link_mapping(
     set_mapping_meter_attr(meta, meter_class, meter_class_override);
 }
 
-#endif /* _DASH_ROUTING_TYPES_P4_ */
\ No newline at end of file
+action set_private_link_nsg_mapping(
```
r12f, line 196: the pl nsg mapping routing type is removed and the tunnel needs to be added in all routing types whenever the tunnel attribute is set.

dash-pipeline/bmv2/dash_metadata.p4
```diff
@@ -187,6 +187,7 @@ struct metadata_t {
     bool dropped;
     encap_data_t encap_data;
     overlay_rewrite_data_t overlay_data;
+    bit<16> tunnel1_id;
```
r12f, line 190: maybe we can simply name it as tunnel id. the reason is that with "direct" routing type, this tunnel will show up as underlay0 instead of underlay1.

dash-pipeline/bmv2/stages/tunnel1.p4
r12f: and once this push_action is moved into each routing type, we can move the tunnel table into the action apply stage.

r12f: and the set tunnel1 attr can be simply push the action and update the tunnel id in the metadata. everything else should be done in the later tunnel action, otherwise the 2 layer of encap won't work, as we only save 1 single layer information in our metadata.

r12f: talking about the bug, I wonder if we could add a test for it : D

dash-pipeline/bmv2/dash_metadata.p4
marian-pritsak: Renamed to dash_tunnel. Tunnel is taken by main SAI.

dash-pipeline/bmv2/stages/tunnel1.p4
marian-pritsak: Moved the tunnel stage after routing_action_apply. Working on a test, will update with a dedicated commit to this PR.

dash-pipeline/bmv2/stages/tunnel1.p4
marian-pritsak: Moved to after routing_action_apply.

dash-pipeline/bmv2/stages/tunnel_stage.p4
```diff
@@ -0,0 +1,41 @@
...
+    inout headers_t hdr,
+    inout metadata_t meta)
+{
+    action set_tunnel_attrs(
+            @SaiVal[type="sai_ip_address_t"]
+            IPv4Address underlay_dip,
+            @SaiVal[type="sai_dash_encapsulation_t", default_value="SAI_DASH_ENCAPSULATION_VXLAN"]
+            dash_encapsulation_t dash_encapsulation,
+            bit<24> tunnel_key) {
+    push_action_static_encap(hdr = hdr,
```
r12f, line 14: Hi Marian, after multiple rounds of thinking, I think it will be better to actually create a dedicated encap data in the metadata struct for tunnel instead of reusing the static encap one. Although the current approach works, but having a dedicated context will make the structure and logic cleaner.

Basically, the proposed update is:
- Create a new encap_data_t field for tunnel.
- Move the tunnel to the pre routing action stage instead of after action apply (which is confusing when looking at the pipeline).

dash-pipeline/bmv2/stages/tunnel_stage.p4
```diff
@@ -0,0 +1,41 @@
+#ifndef _DASH_STAGE_TUNNEL_P4_
+#define _DASH_STAGE_TUNNEL_P4_
+
+control tunnel_stage(
+    inout headers_t hdr,
+    inout metadata_t meta)
+{
+    action set_tunnel_attrs(
+            @SaiVal[type="sai_ip_address_t"]
+            IPv4Address underlay_dip,
```
r12f, line 10: i think it will be better to rename the underlay_dip to dip, because we could have multiple layers of underlay, so it will be a bit confusing, and the dip of the tunnel should be obvious enough anyway.

dash-pipeline/bmv2/stages/tunnel_stage.p4
```diff
@@ -0,0 +1,41 @@
...
+            @SaiVal[type="sai_dash_encapsulation_t", default_value="SAI_DASH_ENCAPSULATION_VXLAN"]
+            dash_encapsulation_t dash_encapsulation,
+            bit<24> tunnel_key) {
```
r12f, line 13: this one doesn't match with the PL NSG HLD tunnel design, so should we update the name?

PR #517, opened 2024-02-14
[dash-p4] Update DASH BM P4 code for fast path SAI API generation

r12f:
This change is to support generating the load balancer fast path SAI APIs. It follows the HLD in the doc repo: <https://github.com/sonic-net/DASH/blob/main/documentation/load-bal-service/fast-path-icmp-flow-redirection.md>.

This change doesn't really contains any real behavior for parsing fast path packets and etc. These updates will come later time.

The updated SAI headers can be previewed here: <https://github.com/r12f/SAI/commit/d0e29049664945a78e3569193d9280fde04654a3>


r12f:
working in progress just in case anyone is interested in preview, will convert it back to PR once it is ready for review.

r12f:
ok, I think it is ready for review now. There are some minor naming changes following what we are doing currently in DASH, so I updated the HLD to make them align.

kcudnik (approved):

marian-pritsak (approved):

r12f:
Thanks Marian! I will submit the SAI header update PR as well today.

PR #520, opened 2024-02-21
[Dash-Sonic] Introduce route group, pa validation tables

prsunny:
1. Add Route group table for grouping routes and binding to an ENI
2. Introduce PA validation table for special validation for certain scenarios

### DIFF
--- a/documentation/general/dash-sonic-hld.md
+++ b/documentation/general/dash-sonic-hld.md
@@ -496,7 +496,7 @@ vm_vni                   = VM VNI that is used for setting direction. Also used 
 ### 3.2.9 ROUTE LPM TABLE - OUTBOUND
 
 ```
-DASH_ENI_ROUTE_TABLE:{{eni}}:
+DASH_ENI_ROUTE_TABLE:{{eni}}
     "group_id": {{group_id}} 
 ```
 
--- a/documentation/general/dash-sonic-hld.md
+++ b/documentation/general/dash-sonic-hld.md
@@ -47,7 +47,7 @@
 |  1.4  | 05/03/2023 |    Prince Sunny     | ACL Tagging, ACL Requirements             |
 |  1.5  | 05/22/2023 | Oleksandr Ivantsiv  | Update configuration examples             |
 |  1.6  | 06/01/2023 |    Prince Sunny     | Added FastPath                            |
-|  1.7  | 02/20/2023 |    Prince Sunny     | Introduce Route Group Table               |
+|  1.7  | 02/20/2024 |    Prince Sunny     | Introduce Route Group Table               |
 
 # About this Manual
 This document provides more detailed design of DASH APIs, DASH orchestration agent, Config and APP DB Schemas and other SONiC buildimage changes required to bring up SONiC image on an appliance card. General DASH HLD can be found at [dash_hld](./dash-high-level-design.md).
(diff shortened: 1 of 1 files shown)


documentation/general/dash-sonic-hld.md
```diff
@@ -494,8 +495,25 @@ vm_vni                   = VM VNI that is used for setting direction. Also used
 
 ### 3.2.9 ROUTE LPM TABLE - OUTBOUND
 
+```
+DASH_ENI_ROUTE_TABLE:{{eni}}:
```
r12f, line 499: nit:
- remove the last ":"
- update eni to "eni_id"

documentation/general/dash-sonic-hld.md
```diff
@@ -494,8 +495,25 @@ vm_vni                   = VM VNI that is used for setting direction. Also used
...
+
+```
+key                      = DASH_ENI_ROUTE_TABLE:eni ; ENI name as key; ENI can bind to a route group table atomically (Overwrites any old bindings)
+; field                  = value 
+group_id                 = Route Group Table Id (contains both IPv4 and IPv6 routes)
+```
+
+```
+DASH_ROUTE_GROUP_TABLE:{{group_id}}
+    "guid": {{string}}
```
r12f, line 511: do we still need the guid here? @sharmasushant

documentation/general/dash-sonic-hld.md
```diff
@@ -1213,7 +1257,20 @@ For the inbound direction, after Route/ACL lookup, pipeline shall use the "under
         "OP": "SET"
     },
     {
-        "DASH_ROUTE_TABLE:F4939FEFC47E:50.1.2.0/24": {
+        "DASH_ENI_ROUTE_TABLE:F4939FEFC47E": {
+	    "group_id":"group_id_2"
+        },
```
kperumalbfn, line 1262: Different ENI ids for group_id_1, group_id_2 and group_id_3.

KrisNey-MSFT:
Currently the route table is modified w/routes belonging to an ENI (per ENI).
Controller calls each route update w/ENI key.
Group table & binding allows for **atomic** binding for an ENI (instead of updating 1 by 1)
(Not a subset of ENI routes).
Cx update their routes in the route table, and expect all or nothing to work.  Adding 1 by 1 we can end up in a non-complete/inconsistent goal state.  

What if libsai has no resources to support an atomic update?  
The controller will need to wait to determine whether this is a successful operation or not.
If unsuccessful, old route table will persist.

documentation/general/dash-sonic-hld.md
prsunny: removed `:`, we use `eni `in other places, so IMO, we can stick to the same format

documentation/general/dash-sonic-hld.md
prsunny: guid can be optional

sharmasushant: I don't see a use for guid field, but I am okay to leave it there.

sharmasushant (approved):
LGTM

KrisNey-MSFT:
Are all comments resolved?  IF Y, I can merge :)

KrisNey-MSFT:
hi @r12f , @kperumalbfn , and @pranjal-git - would you mind reviewing for Prince please?
TY...

documentation/general/dash-sonic-hld.md
prsunny: This is a reference example. it can be any group_id

kperumalbfn (approved):

r12f (approved):

PR #523, opened 2024-02-24
[dash-p4] Create macro for creating direct counters

r12f:
This is to simplify the code for creating and attaching the direct counters.

No SAI updates will be introduced with this change.


r12f:
ouch, seems like we get some infra issue with certs... let me retry the test and see.

image

r12f:
retry works! great!

marian-pritsak (approved):

PR #525, opened 2024-02-26
[SAI] Add dummy buffer and dtel api

kcudnik:
Needed to return dummy success object to for new dash-app


r12f (approved):

KrisNey-MSFT:
Merging per Community call - thank you @kcudnik

PR #526, opened 2024-02-26
Rename ENI attribute PPS to BW to reflect the actual BW limitation in bytes

marian-pritsak:


r12f:
heads to up @prsunny @theasianpianist , this will cause one of the existing SAI attribute being changed. but it should be aligned with our current HLD.

dash-pipeline/bmv2/dash_metadata.p4
```diff
@@ -36,7 +36,7 @@ enum bit<16> dash_tunnel_dscp_mode_t {
 
 struct eni_data_t {
     bit<32> cps;
-    bit<32> pps;
+    bit<32> bw;
```
r12f, line 39: although bw is the one we use in the HLD, I wonder if we like to get the unit clear, such as bw_in_kbps.

chrispsommers: @r12f good idea. At least PPS is unambiguous. The item name should include the units to make it clear.

marian-pritsak: All BW throughout SAI is in Mb/s

KrisNey-MSFT:
Same SAI Challenger issues

KrisNey-MSFT:
@mgheorghe or @chrispsommers - would you be able to help with this one (test failing)?  If N, no worries.

dash-pipeline/bmv2/dash_metadata.p4
mgheorghe: why not keep it `bw` and bw can accept values like 978, 1K, 1M, 15G, 2T, (if we dont want toype too many zeroes) and the last char if letter detonates the unit this way we don't end up with 20 api names, like `bw_in_3_and_a_half_bytes_increments` (exaggeration to make the point)

KrisNey-MSFT:
Lower priority possibly as attribute is not used yet.  BUT, it could likely be merged after fixes, @marian-pritsak thinks it can be done now.  Update SAI headers, need to keep a stable version of SAI headers.  Keep pending for now.

PR #529, opened 2024-02-29
[dash-p4] Remove original overlay IP from encap data and reuse the existing fields.

r12f:
The original overlay dip/sip looks to be redundant and can directly reuse the overlay data we already have in the metadata, hence doing this change.

### DIFF
--- a/dash-pipeline/bmv2/dash_conntrack.p4
+++ b/dash-pipeline/bmv2/dash_conntrack.p4
@@ -48,19 +48,22 @@ control ConntrackIn(inout headers_t hdr,
           }
           restart_expire_timer(); // reset expiration timer for entry
           meta.conntrack_data.allow_in = true;
-          meta.encap_data.original_overlay_sip = original_overlay_sip;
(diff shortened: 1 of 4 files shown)


marian-pritsak (changes requested):
The reason for saving the original SIP is that it may be changed by 4to6 transformation, but we may still need to refer to the original value along the pipeline.

KrisNey-MSFT:
hey @jafingerhut - did this look ok to you (for merge)?

jafingerhut:
@KrisNey-MSFT Executive summary: I am OK with merging this, but that does not provide much higher assurance that these are the changes you want.

Gory details:
For the functional correctness of the changes according to the desired data plane behavior for the DASH project, I am not the best person to review this, unfortunately.

For whether it compiles and runs on P4 DPDK, I believe Chris Sommers has created Github actions/workflow stuff that automatically checks whether it compiles without error, but the various DASH blocking issue for P4 DPDK prevent us from automating checks that pass packets through.  Those checks for BMv2 should both compile and pass packets through on every commit, I believe.

PR #532, opened 2024-03-07
Add DASH HA session API design.

r12f:
This change is made to add HA session API HLD and P4 changes for SAI API generation.

To describe how the HA session API works, the HLD in this change contains:

- The key fundamental components that used to form the HA set topology and their SAI API design.
- How we could implement HA in DASH behavior model and how each components works with each other.
- The life of the packet for inline flow sync.
- The SAI API call sequence of several typical HA related work flows.

The change also contains the P4 code that generates the SAI APIs. And the generated APIs can be previewed in this branch here: <https://github.com/r12f/SAI/tree/user/r12f/ha-api/experimental>.


dash-pipeline/bmv2/stages/ha.p4
```diff
@@ -0,0 +1,133 @@
+#ifndef _DASH_STAGE_HA_P4_
+#define _DASH_STAGE_HA_P4_
+
+control ha_stage(inout headers_t hdr,
```
mukeshmv, line 4: @r12f, as discussed can we add a capability attribute for DPU scope versus ENI scope HA.

r12f: As chatted, I believe it will be better to add it in another PR.

The DPU HA capability will make more sense to be added along with the DPU API or a dedicated one, if needed. Putting the DPU HA capability into ENI HA API will be really weird...

mukeshmv (approved):

dash-pipeline/bmv2/stages/ha.p4
r12f: Actually, let me get some capacities in there, we can probably model this in a generic way.

documentation/high-avail/ha-api-hld.md
```diff
@@ -0,0 +1,617 @@
...
+Whenever a HA set state is changed, it will be reported back via HA set event notification, such as data plane channel goes down. The detailed definition is shown as below:
+
+```c
+/**
+ * @brief HA set event type
+ */
+typedef enum _sai_ha_set_event_t
+{
+    /** Any HA set state is changed, such as data plane channel goes down. */
+    SAI_HA_SET_STATE_CHANGED,
```
marian-pritsak, line 204: This enum should have the possible states to remove the need to read them from the attribute.

documentation/high-avail/ha-api-hld.md
```diff
@@ -0,0 +1,617 @@
...
+ *
+ * @param[in] count Number of notifications
+ * @param[in] data Array of HA set events
+ */
+typedef void (*sai_ha_set_event_notification_fn)(
+        _In_ uint32_t count,
+        _In_ const sai_ha_set_event_data_t *ha_set_event_data);
+```
+
+#### 4.6.2. HA scope event notifications
```
marian-pritsak, line 248: Seems redundant, HA scope does not define liveliness API for this event, it's in the HA_SET.

marian-pritsak (changes requested):

documentation/high-avail/ha-api-hld.md
r12f: ah, yea. i am mostly following what other SAI events is doing, but please correct me if my understanding is not right.

yes, the event enum doesn't contain the detailed state, but we don't need to read them from HA set attribute using query. the reason is the changed attributes will be reported as attr_count and attr in the sai_ha_set_event_data, hence we can directly fetch them there.

I found this approach being a repetitive pattern in existing SAI APIs, so i am mostly following this pattern. It does simplify the event data definition and won't cause state explosion on the enum value.

documentation/high-avail/ha-api-hld.md
r12f: I see, the HA scope state change event is used for reporting the HA state changes. For example, whenever we call SAI API to update the HA role to "Active". It might take some time for the implementation to process the request and really switch to active. Then whenever it is completed, we can send a HA scope event with the HA role set to "Active" in its event data.

This also helps the case where DPU decides to drive the HA state machine without using NPU ([HLD here](https://github.com/r12f/SONiC/blob/user/r12f/ha2/doc/smart-switch/high-availability/smart-switch-ha-dpu-scope-dpu-driven-setup.md)). Whenever the HA role is changed by DPU, it will be reported using this event notification as well.

documentation/high-avail/ha-api-hld.md
marian-pritsak: Notification data doesn't carry attributes but the type that attribute has. See https://github.com/opencomputeproject/SAI/blob/master/inc/saiport.h#L90C5-L90C28 for reference.

documentation/high-avail/ha-api-hld.md
marian-pritsak: Setting HA scope role is synchronous. the set_attribute call should not return until it is completed.

marian-pritsak (changes requested):

documentation/high-avail/ha-api-hld.md
r12f: yea, it could be synchronous for the first case. But for the second case, this notification will be required, otherwise we won't be able to know the state being changed.

documentation/high-avail/ha-api-hld.md
r12f: yea, i know ACL and Port notification is doing it, but it might not be extendable, because we need to change the API whenever a new field needs to be added.

on the otherhand, [FDB](https://github.com/opencomputeproject/SAI/blob/3fe490d836be6ed1d7c4a8763d5e33053f9e957a/inc/saifdb.h#L346), [packet event](https://github.com/opencomputeproject/SAI/blob/3fe490d836be6ed1d7c4a8763d5e33053f9e957a/inc/saihostif.h#L1417) are using attributes directly, which seems to be a more flexiable approach, but i guess the thing I am missing here is the bar, FDB needs to report all its attributes, hence designed like that.

so no problem at all, let me get the event updated.

marian-pritsak: FDB attribute is used to indicate a bridge ID and an entry type - https://github.com/sonic-net/sonic-swss/blob/master/orchagent/fdborch.cpp#L1054, not the state. State is passed as part of the notification data separately. Why would it be harder to extend if both are using the same type?

marian-pritsak (approved):

r12f:
Woot! Thanks @mukeshmv and @marian-pritsak for reviewing! I am merging it now.

PR #533, opened 2024-03-11
Remove INVALID mode from DSCP modes

marian-pritsak:
Preserve is the default, there is no condition under which INVALID can be set.


marian-pritsak:
Need @r12f to approve

r12f (approved):
lgtm!

KrisNey-MSFT:
[Add DSCP mode to DASH 1981](https://github.com/opencomputeproject/SAI/pull/1981/files)

PR #535, opened 2024-03-11
Add ability to provide @validonly tag

marian-pritsak:


marian-pritsak:
Need @r12f to approve

r12f (approved):
LGTM! Thanks Marian!

KrisNey-MSFT:
This is for outer packet.  In Azure there is an option to permit a Cx to set DSCP for VNET.  This should not affect this PR, however want to be cognizant to remember this.

KrisNey-MSFT:
This is outer packet.  We should remember that in Azure we do have scenarios where we permit a Cx to set DSCP in VNET.

PR #537, opened 2024-03-13
[dash-sonic-hld]: Update doc for SONiC DASH virtual switch

Pterosaur:
HLD for SONiC DASH virtual switch


documentation/general/dash-sonic-hld.md
```diff
@@ -1527,3 +1575,61 @@ For the example configuration above, the following is a brief explanation of loo
...
+
+## 4.2 Architecture
+
+![BMv2 virtual SONiC](images/bmv2-virtual-sonic.svg)
+
+## 4.3 Modules
+
+### 4.3.1 BMv2 (dataplane engine)
+
+This component is the original P4 BMv2 container image, which serves as the implementation of the dataplane.
```
r12f, line 1593: "serves as the part of the data plane implementation - usually in hardware."

the word of dataplane is a bit vague, because the app is also part of data plane.

.wordlist.txt
```diff
@@ -720,3 +720,17 @@ Gan
 Ze
 AppDBMemoryEstimation
 Ivantsiv
+SAIRedis
+Dataflow
+POC
+Azp
+cpu
+dpdk
```
r12f, line 728: we can replace all "dpdk" to "DPDK", so we don't need to update the wordlist here.

r12f: same for CPU.

documentation/general/dash-sonic-hld.md
```diff
@@ -1527,3 +1575,61 @@ For the example configuration above, the following is a brief explanation of loo
...
+It attaches three types of interfaces: Ethernet, eth, and cpu.
```
r12f, line 1594: maybe better to give the "cpu" a better name.

r12f: i think we can probably give all these ports better names... the type doesn't have to match with the physical port name. it should give people a better understanding of the architecture when reading the doc.

documentation/general/dash-sonic-hld.md
```diff
@@ -1527,3 +1575,61 @@ For the example configuration above, the following is a brief explanation of loo
...
+- Ethernet is used as the system port. Protocol services like BGP and LLDP perform send/receive operations on these interfaces.
+- eth is used as the line port. These are native interfaces in KVM for communication between the inside and outside. The eth port and Ethernet port is one-to-one mapping.
+- cpu is used for the dpdk port. The dataplane APP directly manipulates the traffic on these ports.
+
+### 4.3.2 Dataplane APP
+
+In the physical DPU, this component is provided by vendors' DPDK application. However, in the virtual switch, our implementation is based on the VPP framework with the CPU interface to simulate it. This dataplane APP loads the generated shared library, saidash.so, which communicates with BMv2 via GRPC. Additionally, this component connects with sairedis through a shim SAI proxy.
```
r12f, line 1601: > In the physical DPU, this component is provided by vendors' DPDK application.

maybe we don't need to mention this in the doc, since how to implement DASH is purely vendor's choice. but to highlight why we are providing this component in our design, e.g. p4 limitation on flow creation, flow resimulation and etc.

r12f: > in the virtual switch

maybe better to change it to "in this virtual DPU", since it is not a switch.

and the same for other places too.

r12f: also we will have a dedicated doc on the data plane app design, hence we can add something like "For more details, please refer to ....." here.

documentation/general/dash-sonic-hld.md
```diff
@@ -1527,3 +1575,61 @@ For the example configuration above, the following is a brief explanation of loo
...
+
+### 4.3.3 SAIRedis
+
+In the original virtual SONiC, SAIRedis will load the saivs.so. However, in the SONiC DASH virtual switch, it will load the sai proxy mentioned above.
```
r12f, line 1605: maybe change "sai proxy" to "remote sai client" to match with the graph. in case, the word "proxy" causes any confusion.

documentation/general/dash-sonic-hld.md
```diff
@@ -1527,3 +1575,61 @@ For the example configuration above, the following is a brief explanation of loo
...
+
+### 4.3.4 SWSS
+
+The SWSS on the virtual switch is almost the same as the one used in the physical switch (DPU). We don't need to make any special changes to it. However, due to some tedious historic issues, the SWSS on the virtual switch will make some extra SAI calls (like DTEL). To adapt to these calls, we will generate a corresponding mock SAI implementation in saidash.so.
```
r12f, line 1609: > We don't need to make any special changes to it. However, due to some tedious historic issues, the SWSS on the virtual switch will make some extra SAI calls (like DTEL). To adapt to these calls, we will generate a corresponding mock SAI implementation in saidash.so.

Since you have mentioned the .so files in the doc, do you mind to update the graph to have these .so files captured too?

r12f: > However, due to some tedious historic issues, the SWSS on the virtual switch will make some extra SAI calls (like DTEL). To adapt to these calls, we will generate a corresponding mock SAI implementation in saidash.so.

For this part, we can move it to SAIRedis part, and rephrase it to "For the SAI APIs that will not be used by DASH/DPU SONiC, the SAI implementation will be mocked, as long as SWSS could work, e.g. DTEL."

documentation/general/dash-sonic-hld.md
```diff
@@ -1527,3 +1575,61 @@ For the example configuration above, the following is a brief explanation of loo
...
+
+### 4.3.5 Underlay Service
```
r12f, line 1611: Maybe we can rename this section to "Other SONiC services". And these services will be kept, so the KVM runs the same way as how SONiC runs on the real DPU.

documentation/general/dash-sonic-hld.md
```diff
@@ -1527,3 +1575,61 @@ For the example configuration above, the following is a brief explanation of loo
...
+
+We plan to keep the underlay services, such as BGP, LLDP, and others. This is because we not only want to simulate a DPU but also provide an opportunity to use a virtual SONiC image to simulate a smart switch with a single DPU for simplified scenarios.
+
+## 4.4 Dataflow
+### 4.4.1 Data plane
+
+All data plane traffic will enter the BMv2 simple switch and be forwarded to the target port based on the P4 logic imported on BMv2.
+
+```text
+E.G.
```
r12f, line 1621: mermaid sequence graph will be greatly helpful : D

documentation/general/dash-sonic-hld.md
```diff
@@ -1527,3 +1575,61 @@ For the example configuration above, the following is a brief explanation of loo
...
+VNet packet:
+  eth1<->dash pipeline(BMv2)<->eth2
+
+TCP SYN packet:
+  eth1<->cpu0<->dataplane APP<->cpu0<->eth2
+```
+
+### 4.4.2 Control plane
+
+In the physical SmartSwitch, configuration is forwarded via GNMI in the NPU. So, in the virtual SONiC environment, the SWSS module is capable of receiving configuration from an external GNMI service through the management port, eth-midplane. However, in the single device mode, the GNMI service can also be run within the KVM and directly forward the configuration to the local SWSS module.
```
r12f, line 1635: the gNMI is actually missing in the architecture graph, do you mind to add it there? I think it is way more important than BGP, since we are not using BGP on DPU anyway in SmartSwitch setup.

documentation/general/dash-sonic-hld.md
```diff
@@ -1527,3 +1575,61 @@ For the example configuration above, the following is a brief explanation of loo
...
+## 4.4 Dataflow
+### 4.4.1 Data plane
+
+All data plane traffic will enter the BMv2 simple switch and be forwarded to the target port based on the P4 logic imported on BMv2.
+
+```text
+E.G.
+
+LLDP packet:
+  eth1<->Ethernet0<->LLDP process
```
ashutosh-agrawal, line 1624: This will require bmv2 implementation to trap LLDP and other control packets to the CPU but this functionality is not part of DASH.p4. Also, in the HW implementation, control packets will be trapped by NPU and not DPU.

Pterosaur: I agree that may not be a part of DASH.p4 but I feel it's a part of DPU.
If my understanding is correct, the NPU and DPU will be connected via Ethernet port and the LLDP traffic will also be running between them.

documentation/general/dash-sonic-hld.md
```diff
@@ -1,31 +1,78 @@
 # SONiC-DASH HLD
 ## High Level Design Document
-### Rev 1.7
+### Rev 1.8
```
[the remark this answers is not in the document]
Pterosaur: Move this PR to: https://github.com/sonic-net/SONiC/pull/1634

Pterosaur:
Move this PR to: https://github.com/sonic-net/SONiC/pull/1634

documentation/general/dash-sonic-hld.md
```diff
@@ -1527,3 +1575,61 @@ For the example configuration above, the following is a brief explanation of loo
 		h. Third Action is Appliance Encap for id 22
 		i. Packet shall be encapsulated with Outer DIP as 100.8.1.2 and SIP as VIP of this originating appliance card with VNI of 101. 
 		j. Inbound flow shall be similar to PL and outer encap shall be of the SLB MUX and not of the NSG appliance.
+
+# 4 SONiC DASH virtual switch
```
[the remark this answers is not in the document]
Pterosaur: Done

documentation/general/dash-sonic-hld.md
Pterosaur: Done

.wordlist.txt
Pterosaur: Done

documentation/general/dash-sonic-hld.md
Pterosaur: Done

documentation/general/dash-sonic-hld.md
Pterosaur: Done in the new PR

documentation/general/dash-sonic-hld.md
Pterosaur: Done

documentation/general/dash-sonic-hld.md
Pterosaur: I checked that the .so name should be libsai.so instead of saidash.so. So I removed the `.so` and jsut call it saidash. And Move it into Dataplane APP section. Because these mock API is implemented in saidsah instead of the shim proxy layer.

documentation/general/dash-sonic-hld.md
Pterosaur: Done

documentation/general/dash-sonic-hld.md
Pterosaur: Good idea, Done.

documentation/general/dash-sonic-hld.md
Pterosaur: Done

Pterosaur: Done

PR #546, opened 2024-04-02
Add routing groups

marian-pritsak:
Make an intermediate stage between ENI and routing to be able to atomically bind/unbind an entire
LPM table to/from ENI.


dash-pipeline/bmv2/stages/outbound_routing.p4
```diff
@@ -6,12 +6,29 @@
 control outbound_routing_stage(inout headers_t hdr,
                                inout metadata_t meta)
 {
+
+    action set_routing_group_version(bit<1> admin_state) {
```
r12f, line 10: since the parameter is not a version, do you mind to change this to `set_routing_group_attr`?

dash-pipeline/bmv2/stages/outbound_routing.p4
```diff
@@ -33,8 +50,13 @@ control outbound_routing_stage(inout headers_t hdr,
             return;
         }
 
+        routing_group.apply();
+        if (meta.eni_data.routing_group_admin_state) {
```
r12f, line 54: maybe it is better to create a dedicated structure for routing group data, then we can use meta.routing_group_data.admin_state to access the related states. 

As metadata becomes more and more complicated, it will be more future-proof.

dash-pipeline/bmv2/stages/outbound_routing.p4
```diff
@@ -33,8 +50,13 @@ control outbound_routing_stage(inout headers_t hdr,
...
         routing.apply();
+        } else {
```
r12f, line 56: nit: drop and return first might be better then if else : D

dash-pipeline/bmv2/stages/outbound_routing.p4
r12f: also do you mind to change admin_state to disabled?

there are 2 reasons:
- disabled is more intuitive than admin_state.
- admin_state should not be frequently used (otherwise, why programming it), hence it will be better to make use of the default value. 

Making this field disabled with default value set to false seems to be a good way to achieve these 2 things.

dash-pipeline/bmv2/stages/outbound_routing.p4
```diff
@@ -33,8 +50,13 @@ control outbound_routing_stage(inout headers_t hdr,
...
+            drop(meta);
```
r12f, line 57: if the packet can be dropped due to admin state down, it will be better to add a counter on top of routing group as stats.

KrisNey-MSFT:
PTF tests ok, SAI Challenger tests will be run locally on Marian's machine

KrisNey-MSFT:
DPUGen to generate config; has hard-coded entries; ENI_ID is no longer a field; @mgheorghe will take a look at fixing the test, fix the outbound routing entry to use routing_group ID and struct instead of ENI.

dash-pipeline/bmv2/stages/outbound_routing.p4
marian-pritsak: Done

dash-pipeline/bmv2/stages/outbound_routing.p4
marian-pritsak: Done

dash-pipeline/bmv2/stages/outbound_routing.p4
marian-pritsak: Done

dash-pipeline/bmv2/stages/outbound_routing.p4
marian-pritsak: Done

r12f (approved):

PR #548, opened 2024-04-08
[doc] Add bulk sync support in HA session API.

r12f:
The initial HA session API added the support of HA state machine management and support of configurating data plane channel for inline sync. And this change added the API to support bulk sync.

From SAI point of view, the bulk sync works as below:

1. On the source side, a flow API will be called to create the bulk sync session with specified conditions, which points to a certain destination, which can either be the swbusd in the control plane for flow relay or the paired DPU directly.
2. Once session is created, the source DPU will start to connect to the destination w/ a gRPC server and send the flow record across.
3. Once session is done, optionally the connection can be closed.

Hence, in the API, we will see 2 major things being added:

- On HA set attributes, the control plane channel port will be added.
- More stats will be added on the HA set to show how the flow sync works.


mukeshmv (approved):

r12f:
@marian-pritsak , please let us know if you have any new concerns. We will first move on first and adjust later with your comments.

PR #549, opened 2024-04-09
[doc] Add flow reconcile APIs in the HA session API.

r12f:
The flow reconcile API is for helping unplanned failover to avoid SDN policy going backwards. 

Basically, whenever the DPU/ENI enters a role that does flow decision from a role that doesn't do it. It needs to stop flow resimulation for existing flows, even upon request. Until the flow reconcile request is called explicitly, when SDN controller confirms the latest policies are programmed, the flow resimulation can be resumed.

Because the flow reconcile goes with the DPU/ENI, the attributes are added on the HA scope level. Besides attributes, we also added an event notification for requesting flow reconcile from the DASH implementation.


mukeshmv (approved):
lgtm

PR #550, opened 2024-04-10
[dash-p4] Update DASH p4 for support the latest metering design.

r12f:
To support the latest metering design, we are updating the DASH p4 code to generate the SAI header file as well as updating how the metering pipeline works.

- The latest metering HLD is shown here: https://github.com/sonic-net/DASH/pull/544
- The SAI API preview branch can be found here: https://github.com/r12f/SAI/tree/user/r12f/meter


dash-pipeline/bmv2/stages/metering_update.p4
```diff
@@ -82,27 +80,19 @@ control metering_update_stage(
...
-
         meter_bucket.apply();
+
         if (meta.direction == dash_direction_t.OUTBOUND) {
-            UPDATE_COUNTER(meter_bucket_outbound, meta.meter_bucket_index);
+            UPDATE_COUNTER(meter_bucket_outbound, meta.meter_class);
         } else if (meta.direction == dash_direction_t.INBOUND) {
-            UPDATE_COUNTER(meter_bucket_inbound, meta.meter_bucket_index);
+            UPDATE_COUNTER(meter_bucket_inbound, meta.meter_class);
         }
```
marian-pritsak, line 96: We should also check if meter class is not 0 for this section of code.

marian-pritsak (changes requested):

dash-pipeline/bmv2/stages/metering_update.p4
r12f: You are right! This is a good one!

r12f: It is updated now. : D

marian-pritsak (approved):

r12f:
thanks a lot Marian!

PR #551, opened 2024-04-10
[build] Fix SAI challenger build break.

r12f:
## Problem

We are seeing a problems in our repo. Some recent changes in our dependency caused a build break due to a commit id being removed as below.

> ...
> Fetching submodule usecases/sai-ptf/SAI
> fatal: remote error: upload-pack: not our ref a249f0c0cf70c954fdd05121bbc1913185afb713
> fatal: the remote end hung up unexpectedly
> Fetched in submodule path '../test/SAI-Challenger', but it did not contain ccff50f4ae7371b7ece63ecab9866ee0f617ea45. Direct fetching of that commit failed.
> make: *** [Makefile:86: sai-submodule] Error 1

To fix the SAI-Challenger build problem, we cannot simply rollback or move to latest master, as we need some new changes there.

So this change will put a bandit on the SAI-Challenger, so we can temporarily workaround this issue, and after the build is fixed, we will update the SAI-Challenger again to the latest master with all the features being ready there.


ashutosh-agrawal (approved):

r12f:
thanks a lot @ashutosh-agrawal !

PR #553, opened 2024-04-18
[dash-p4] Add support to generate flow resimulation APIs

r12f:
This change updates the P4 pipeline to generate the flow resimulation APIs.

The change is following the DASH flow resimulation HLD defined in the PR here: https://github.com/sonic-net/DASH/pull/543.

The flow resimulation could not be directly implemented using plain p4 api call. The actual pipeline implementation will be postponed until data plane app is ready.

For the generated SAI API, the routing actions type will be generated as flag defined as below:

image


KrisNey-MSFT:
quick ping to @marian-pritsak

marian-pritsak (approved):

PR #554, opened 2024-04-19
[sai-gen] Simplify condition for table action generation.

r12f:
## Problem

Currently, we have a very complicated check for when a table action should be generated, for example:

image

The reason of this check being introduced is to enable us ignore the `SAI_VIP_ENTRY_ACTION_DENY` and `SAI_PA_VALIDATION_ENTRY_ACTION_DENY` action while still keep the action generated for them, as these entries only has 1 action which will be ignored.

However, this check is too narrowed, and also create some issues, for example - an object (not entry) with 1 action will also have the action type generated, which is not ideal.

Ideally, the condition to generate the action type should be really straightforward:
1. If the table entry has more than 2 valid actions (not deny), the action type should be generated.
2. If the table entry is an entry, not object, the action type should be generated.
3. Otherwise, the action type should not be generated.

This simple logic simplify the mind model a lot, and the only case it didn't cover is the eni_eth_address_map_entry, which IMO, should follow the same pattern as many other similar tables (direction lookup, VIP, PA validation) and have the action type created too.

image

Hence, proposing the change to make our life easier.

### DIFF
--- a/dash-pipeline/SAI/templates/saiapi.h.j2
+++ b/dash-pipeline/SAI/templates/saiapi.h.j2
@@ -34,7 +34,7 @@
  */
 
 {% for table in sai_api.tables %}
-{% if table.actions | length > 1 or ((table.actions | length == 1) and (((table.is_object == 'false') or (table['keys'] | length <= 1)) and ((table['action_params'] | length == 0))))  %}
+{% if table.actions | length > 1 or table.is_object == 'false' %}
 /**
  * @brief Attribute data for #SAI_{{ table.name | upper }}_ATTR_ACTION
  */
@@ -103,7 +103,7 @@ typedef enum _sai_{{ table.name }}_attr_t
     SAI_{{ table.name | upper }}_ATTR_START,
 
 {% set ns = namespace(firstattr=false) %}
-{% if table.actions | length > 1 or ((table.actions | length == 1) and (((table.is_object == 'false') or (table['keys'] | length <= 1)) and ((table['action_params'] | length == 0))))  %}
+{% if table.actions | length > 1 or table.is_object == 'false'  %}
     /**
      * @brief Action
      *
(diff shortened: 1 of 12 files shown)


dash-pipeline/SAI/templates/saiapi.h.j2
```diff
@@ -34,7 +34,7 @@
  */
 
 {% for table in sai_api.tables %}
-{% if table.actions | length > 1 or ((table.actions | length == 1) and (((table.is_object == 'false') or (table['keys'] | length <= 1)) and ((table['action_params'] | length == 0))))  %}
+{% if table.actions | length > 1 or table.is_object == 'false' %}
```
mukeshmv, line 37: Original check tries to generate action-type even for sai object if it has only 1 action with 0 params. Also it skips 0 action case. If these are no longer needed / not possible, then this simpler should be fine.

r12f: yea, there are 2 cases here:

- Case 1: Only 1 action with 0 parameters - this is added to cover the VIP and PA validation case for permit action, which doesn't have any parameters. Both VIP and PA validation are entries not objects, so `is_object == false` will get them covered there.
- Case 2: 0 action case. This can only happen if a P4 table has only 1 defaultonly action. In this case, it is better to directly apply the action instead of introducing a table.

so to me, it looks these two cases are both ok to be simplified. what do you think?

mukeshmv (approved):

dash-pipeline/SAI/templates/saiapi.h.j2
mukeshmv: sure

r12f:
thanks Mukesh for reviewing the change!

dash-pipeline/SAI/templates/saiapi.h.j2
r12f: also for the case 1, 0 parameter is ok, as we can see after the check is removed, the SAI API generation works and generated API looks the same.

r12f: thanks Mukesh. as you approved this PR, I am resolving the comment now.

PR #556, opened 2024-04-22
Add tunnel stage

marian-pritsak:
Add an extra tunnel stage for a second layer
of encapsulation


dash-pipeline/bmv2/stages/tunnel_stage.p4
```diff
@@ -0,0 +1,41 @@
+#ifndef _DASH_STAGE_TUNNEL_P4_
+#define _DASH_STAGE_TUNNEL_P4_
+
+control tunnel_stage(
+    inout headers_t hdr,
+    inout metadata_t meta)
+{
+    action set_tunnel_attrs(
+            @SaiVal[type="sai_ip_address_t"]
+            IPv4Address underlay_dip,
```
r12f, line 10: I believe `dip` is good enough here, as the tunnel can be multiple tunnels, using underlay can be misleading.

dash-pipeline/bmv2/stages/tunnel_stage.p4
```diff
@@ -0,0 +1,41 @@
...
+            @SaiVal[type="sai_dash_encapsulation_t", default_value="SAI_DASH_ENCAPSULATION_VXLAN"]
+            dash_encapsulation_t dash_encapsulation,
+            bit<24> tunnel_key) {
+    push_action_static_encap(hdr = hdr,
+                            meta = meta,
+                            encap = dash_encapsulation,
+                            vni = tunnel_key,
+                            underlay_sip = hdr.u0_ipv4.src_addr,
+                            underlay_dip = underlay_dip,
+                            overlay_dmac = hdr.u0_ethernet.dst_addr);
```
r12f, line 20: should we use the meta.overlay_data.dmac here, since it could be changed in the middle of the pipeline.

dash-pipeline/bmv2/dash_pipeline.p4
```diff
@@ -353,6 +354,8 @@ control dash_ingress(
 
         routing_action_apply.apply(hdr, meta);
 
+        tunnel_stage.apply(hdr, meta);
```
r12f, line 357: do you mind to move the tunnel stage into routing_action_apply stage?

This helps maintaining the pipeline in mind model (once action_apply is done, packet transformation is done other than underlay routing).

dash-pipeline/bmv2/stages/tunnel_stage.p4
marian-pritsak: At this point, it's the same value already

dash-pipeline/bmv2/stages/tunnel_stage.p4
marian-pritsak: Done

dash-pipeline/bmv2/dash_pipeline.p4
marian-pritsak: Done

dash-pipeline/bmv2/stages/tunnel_stage.p4
```diff
@@ -0,0 +1,41 @@
...
+        }
+
+        actions = {
+            set_tunnel_attrs;
+        }
+    }
+
+    apply {
+        if (tunnel.apply().hit) {
+            do_action_static_encap.apply(hdr, meta);
```
r12f, line 36: do you mind to make the tunnel using another encap data instead of the static encap one?

this makes the data plane app hard to implement, because the tunnel info overrides the static encap, which makes us impossible to create the flow.

r12f: once we have this changed, we can move the tunnel stage to pre_action_apply stage, which makes the pipeline more clear - action apply should be the last thing we do before underlay routing.

dash-pipeline/bmv2/stages/routing_action_apply.p4
```diff
@@ -15,7 +16,9 @@ control routing_action_apply(
         // because it requires the transforms on the inner packet to be finished in order to
         // get the correct inner packet size and other informations.
         do_action_static_encap.apply(hdr, meta);
+
+        tunnel_stage.apply(hdr, meta);
```
r12f, line 20: this should happen before action apply stage.

dash-pipeline/bmv2/stages/tunnel_stage.p4
marian-pritsak: Decoupled tunnel_encap from static_encap routing action

dash-pipeline/bmv2/stages/routing_action_apply.p4
marian-pritsak: It does not makes sense. Tunnel stage works on outer layers. If I move it before static_encap, it will mess up the tunnel_pointer.

r12f: yea, sorry, I probably didn't make it clear. 

In my mind, the "tunnel stage" is the tunnel id to tunnel lookup, just like the routing stage and mapping stage, so the lookup needs to happen before the static encap.

However, like you mentioned, applying tunnel encap for changing the packet should happen after the static encap, and this keeps the tunnel pointer working.

marian-pritsak: Fixed, split lookup from the encap itself.

r12f (approved):

PR #558, opened 2024-04-24
Fix entry action search in saiapi.cpp.j2

marian-pritsak:
The refactoring introduced in https://github.com/sonic-net/DASH/pull/438 only looks for a proper action attribute in SAI objects, not entries. In case of SAI entry, it iterates through all the attributes and matches the values regardless of attribute ID.

This PR aligns action search for both objects and entries, and adds action ID attributes to missing create_entry calls in tests.


r12f:
hi Marian, do you mind to summary a bit on the background of this PR?

marian-pritsak:
> hi Marian, do you mind to summary a bit on the background of this PR?

Done.

r12f (approved):
lgtm! thanks Marian!

dash-pipeline/SAI/templates/saiapi.cpp.j2
```diff
@@ -443,22 +443,25 @@ static sai_status_t dash_sai_create_{{ table.name }}(
     // Search the action
     for (uint32_t i = 0; i < attr_count; i++)
     {
-        switch(attr_list[i].value.s32)
+        if (SAI_{{ table.name | upper }}_ATTR_ACTION == attr_list[i].id)
```
kcudnik, line 446: previously there was switch, and now there is if, and if attribure is not action, then all other attributes are skipped, in previous case, at least it was logged that some attribue is passed, now there is no log message, 

i propose to make this (if) and then at the end make continue, and leave switch statement as is after if.ACTION case

PR #562, opened 2024-05-09
[sai] Update SAI to latest master

r12f:
We have the metering API, flow resimulation API, HA state management and inline sync APIs, and more updates all merged into OCP SAI. This change brings the updates back to DASH, so we can see the latest differences.


mukeshmv (approved):
lgtm

r12f:
thanks Mukesh for the review!

r12f:
The SAI thrift issue is fixed and merged into SAI master, and the commit ID is updated to point to it.

PR #563, opened 2024-05-10
[doc] Include drop counters in HA API.

r12f:
Packet drops is one of the major common problem in network that needs to be monitored. This change adds the drop counters that we need to monitor in DASH pipeline.

Since the drop counters are not too complicated and also could be counted as part of the HA design, hence we are adding them as part of the HA API doc.


marian-pritsak (approved):

r12f:
thanks Marian for the review!

PR #564, opened 2024-05-10
[dash-p4] Support SAI API generation for bulk sync API, flow reconcile API and related counter.

r12f:
This change updates the P4 code to generate the SAI APIs for bulk sync API, flow reconcile API and related counters.

The API change can be found via the PR in OCP SAI repo here: https://github.com/opencomputeproject/SAI/pull/2014.


mukeshmv (commented):
lgtm

r12f:
hi @marian-pritsak , please let us know if you have any comments after you are back. we will get the HLD adjusted accordingly afterwards.

PR #565, opened 2024-05-19
Skip default IPv6 route with prefix zero and fix IPv6 prefix bug

Pterosaur:
1.To set a default IPv6 route with the prefix 0, we will get following error logs. It's caused by the limitation of p4lang: https://github.com/p4lang/PI/blob/24e0a3c08c964e36d235973556b90e0ae922b894/proto/frontend/src/device_mgr.cpp#L2242-L2246 .

```
2024 Jun  3 15:26:21.033130 vlab-01 ERR syncd#syncd_dash: e:- mutateTableEntry: GRPC ERROR[2]: , #010#002#032¡#001#012#037type.googleapis.com/p4.v1.Error#022~#010#003#022gInvalid reprsentation of 'don't care' LPM match, omit match field instead of using a prefix length of 0#032#021ALL-sswitch-p4org
2024 Jun  3 15:26:21.033130 vlab-01 ERR syncd#syncd_dash: e:- mutateTableEntry: GRPC call Write::INSERT ERROR: table_id: 49279256 match { field_id: 1 lpm { value: "\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000" } } action { action { action_id: 32404057 params { param_id: 1 value: "\000\000" } params { param_id: 2 value: "\000\000" } } }
```
In this case, I just ignore the default IPv6 route to skip this limitation.

2. Fix IPv6 prefix length function bug.
There are two issues on the IPv6 prefix:
- when we calculate the prefix length from the mask, we should consider the trailing zeros from high address space, aka, the end of mask array, of SAI IPv6 attribution. So, I revised this issue by reversing the IPv6 mask array.
- Because the function `leadingNonZeroBits` returns the exact length of prefix, we shouldn't directly subtract it from 129. For example, if the mask is 0xffffffffffffffffffffffff, the original implementation will get 129-0-32=97, it is obviously wrong. The correct answer is 128. So, I fix this bug.

### DIFF
--- a/dash-pipeline/SAI/src/utils.h
+++ b/dash-pipeline/SAI/src/utils.h
@@ -27,7 +27,6 @@ extern "C"
 #include <limits>
 #include <fstream>
 #include <memory>
-#include <type_traits>
 
 namespace dash
 {
@@ -240,13 +239,13 @@ namespace dash
         int leadingNonZeroBits(const sai_ip6_t& ipv6);
 
         template<typename T>
-            sai_status_t ipPrefixSetVal(const sai_attribute_value_t &value, T &t, int bits = -1)
+            void ipPrefixSetVal(const sai_attribute_value_t &value, T &t, int bits = -1)
             {
-                return ipPrefixSetVal(value.ipprefix, t);
+                ipPrefixSetVal(value.ipprefix, t);
             }
 
         template<typename T>
-            sai_status_t ipPrefixSetVal(const sai_ip_prefix_t &value, T &t, int bits = -1)
+            void ipPrefixSetVal(const sai_ip_prefix_t &value, T &t, int bits = -1)
             {
                 switch(value.addr_family)
                 {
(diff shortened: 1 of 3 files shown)


r12f:
Hi @Pterosaur , do you mind to add some description here to describe the issue and how we are fixing it?

dash-pipeline/SAI/src/utils.h
```diff
@@ -268,6 +269,10 @@ namespace dash
                     case SAI_IP_ADDR_FAMILY_IPV6:
                         {
                             uint8_t ip[16];
+                            if (std::is_same<T, p4::v1::FieldMatch_LPM>::value) {
+                                // BMv2 cannot support IPv6 LPM with prefix
+                                return SAI_STATUS_NOT_SUPPORTED;
```
r12f, line 274: is this the BMv2's limitation or it is our implementation limit?

also i am a bit hesitated that we should check and fail here, because of 2 reasons:
- It is a utility function and should not have any knowledge on the scenarios
- Also the type check is a bit hacky.

IMO, if this is something we don't support, the change in the template should be good enough.

Pterosaur:
> Hi @Pterosaur , do you mind to add some description here to describe the issue and how we are fixing it?

Hi @r12f , Thanks for your suggestion. This PR is just for testing right now, So it looks hacky and drafted. I will refactor it after confirmed.

Pterosaur:
> hi Ze, do you mind to add some explanation on what bug we are fixing in this change and how?

@r12f Thanks for your suggestion, Please check it again.

KrisNey-MSFT:
@r12f - looks like a request for you to re-check after edits :)

r12f (approved):

KrisNey-MSFT:
This is to ensure bmv2 can launch, but is a temporary patch.  Not valid in LPM, and it fails.  
@r12f will designate someone to file an Issue in **P4 Runtime GitHub** (leading 0's) if we feel it is related here. 
We need a volunteer to fix this if we can figure it out.

PR #568, opened 2024-05-25
[dash-p4] Refactor global and ENI counter definitions.

r12f:
Moving the global and ENI counters from all over the places to a dedicated file, because they are special and working more like "global" counters, and spread across the pipeline.

This helps us to have a better view and helps us maintaining the ABI compatibility of these counters.

The other counters will go with its table, such as HA set, because they are only used in logic that is related to that single table.


r12f:
Waiting for the drop counter change to be checked in.

r12f:
The drop counter P4 update is merged. This PR is ready for review.

PR #569, opened 2024-05-26
[dash-p4] Refactor incoming packet encap handling and inbound routing stage.

r12f:
Currently, several stages / tables in DASH directly uses the headers to lookup entries, such as direction. inbound routing, and etc. It it ok currently, because DASH only accepts traffic with a single layer and a single type of encap. However, once we start to handle multiple layers and types of encaps, it will make the pipeline very complicated.

This change refactors the pipeline to add a pre-pipeline stage to preprocess the underlay and overlay data for pipeline lookup and actions, so we can keep the rest of the pipeline more future-proof.

### DIFF
--- a/dash-pipeline/SAI/specs/dash_appliance.yaml
+++ b/dash-pipeline/SAI/specs/dash_appliance.yaml
@@ -27,9 +27,9 @@ sai_apis:
   p4_meta: !!python/object:utils.sai_spec.sai_api_p4_meta.SaiApiP4Meta
     tables:
     - !!python/object:utils.sai_spec.sai_api_p4_meta.SaiApiP4MetaTable
-      id: 42701762
+      id: 41082190
       actions:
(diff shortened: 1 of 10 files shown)


KrisNey-MSFT:
Does this need a reviewer @r12f ?  Let me know if I can help...

r12f:
yea, I have added Marian in the PR. Let's see how it goes.

r12f:
Thanks Kristina.

marian-pritsak (approved):
Looks clean with regards to API.

KrisNey-MSFT:
@r12f  - looks good to reviewer, if you'd like to take a look

KrisNey-MSFT:
@r12f - looks like this has branch conflicts now - we will wait for you to return from your OOF.

KrisNey-MSFT:
checking in here @r12f

PR #570, opened 2024-05-26
[dash-sai] Update SAI API template to make SAI explorer working better.

r12f:
There is a change directly committed to OCP SAI to make the DASH SAI APIs working better for SAI explorer. However, this change is not updated into DASH repo, which means all future SAI API updates from DASH could override it and any new APIs will not be covered.

To help with this, this change updates the SAI API template to have a better naming.

For references, the original OCP PR is here: https://github.com/opencomputeproject/SAI/pull/1971.


mukeshmv (approved):
lgtm

r12f:
thanks a lot for the review, Mukesh!

PR #571, opened 2024-05-26
[sai-gen] Refactor sai_api_gen.py into python modules.

r12f:
As the SAI generator becomes more and more complicated, it makes it hard to maintain with a single python file. Hence, for better readability, refactor the single file into python modules, with almost 1 class per 1 file.

This change is a pure code move, without changing any functionalities.


mukeshmv (approved):
lgtm

KrisNey-MSFT:
There is a bug where Marian found a workaround re: reduction logic for action parameters that repeat across different actions. Ex: Underlay_DIP appearing in multiple actions will be presented as 1 attribute in SAI API. Only 1 instance of parameter will be carried w/param_id (not same across actions). Riff is aware...refactoring _should not_ break...

PR #572, opened 2024-05-27
[sai-gen] Rename SAI generator SAI* objects to DashP4* objects.

r12f:
This change is purely for renaming, without any functionality or logic change. It rename SAI* objects to DashP4* objects, for example, SAIEnum to DashP4Enum.

The reason of this change is to get us prepared for introducing the SAI spec objects for SAI API generation. It allows us to separate the P4Runtime parsing logic with the real SAI API definitions, as well as greatly simplify the logic in the j2 templates, which is extremely hard to maintain.


KrisNey-MSFT:
@mukeshmv OR @marian-pritsak - could you take a look at this minor change?

mukeshmv (approved):
lgtm

PR #576, opened 2024-06-01
[sai-gen] Parse P4 metadata into SAI spec for lib generation.

r12f:
The current SAI spec has all the information for header generation. To simplify the lib generation as well, parse the SAI to P4 info mapping into SAI spec too.

This can simplify the SAI lib templates, as we don't have to do too many checks in the templates anymore.


r12f:
Hi @marian-pritsak check out the ca to pa spec! : D with this info, we should be able to fix the field id problem for good!

r12f:
Splitted this change into 2: this one and #577 . Since this change is dependent on #577 , I am making this one a draft for now.

marian-pritsak (approved):

PR #577, opened 2024-06-02
[sai-gen] Support merging SAI specs.

r12f:
This change added the SAI spec merging support. This is to ensure any new SAI attribute changes will follow the rules below to help ABI compatibility:

- All new attributes or action enum values will be added in the end of the list.
- Existing attributes will be updated inline. In the future, we can add more check here, e.g., type changed in non-compatible way and etc.
- All old attributes will be marked as deprecated instead of removed.

Recently, the tunnel APIs are added, but because it merged with the SAI spec PR together without rebase, it is not captured in the previous SAI spec. With SAI spec merging, we can see the changes are showing up in the expected sequence. This would be an example that demos the several usages of SAI spec.

### DIFF
--- a/dash-pipeline/SAI/specs/dash_tunnel.yaml
+++ b/dash-pipeline/SAI/specs/dash_tunnel.yaml
@@ -1,0 +1,48 @@
+!!python/object:utils.sai_spec.sai_api_group.SaiApiGroup
+name: dash_tunnel
+description: ''
+sai_apis:
+- !!python/object:utils.sai_spec.sai_api.SaiApi
+  name: dash_tunnel
+  description: ''
+  is_object: true
+  enums: []
+  structs: []
+  attributes:
+  - !!python/object:utils.sai_spec.sai_attribute.SaiAttribute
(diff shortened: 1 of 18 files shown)


dash-pipeline/SAI/specs/dash_tunnel.yaml
```diff
@@ -0,0 +1,48 @@
...
+    attr_value_field: s32
+    default: SAI_DASH_ENCAPSULATION_VXLAN
+    isresourcetype: false
+    flags: CREATE_AND_SET
+    object_name: null
+    allow_null: false
+    valid_only: null
+    deprecated: null
+  - !!python/object:utils.sai_spec.sai_attribute.SaiAttribute
+    name: SAI_DASH_TUNNEL_TUNNEL_KEY
```
r12f, line 37: Hi @marian-pritsak , according to HLD, this one should be VNI. I wonder if you could help get it updated?

KrisNey-MSFT:
Request to @marian-pritsak for assistance

marian-pritsak (approved):

PR #578, opened 2024-06-02
[sai-gen] Update SAI API header file generation to use SAI spec

r12f:
Moving SAI API header file generation from using the P4 table information to SAI spec.

This change contains several minor fixes to the SAI specs:
- Description and comment generation with respect to the SAI acronyms.
- Fixed SAI flags and default values.
- Added missing SAI attributes that is generated from table keys.
- Fixed valid only tag generation.
- Fixed unwanted header files for underlay.

Some typical changes (all comment changes, APIs are not changed and kept in the same order):

image

image


r12f:
adding Ze into review, since his DPU image work is getting affected if the SAI API is not backward compatible.

r12f:
without this change, the header generation will remain broken.

Pterosaur (approved):

PR #588, opened 2024-06-11
Add missing masks to PL action

marian-pritsak:


r12f:
Hi Marian, do you mind to have the PTF test also planned for this work?

r12f (approved):

marian-pritsak:
> Hi Marian, do you mind to have the PTF test also planned for this work?

Yes, I have a few in mind - for each routing type, but I'll need to fix the param order first because it will cause all sorts of issues.

PR #589, opened 2024-06-13
[dash-sai] Update object id manager to support extensions range

kcudnik:
Update OID format to support new extensions range base More info in PR: https://github.com/opencomputeproject/SAI/pull/2028


r12f (approved):
lgtm! thanks Kamil!

PR #590, opened 2024-06-13
Update dash_acl.p4 Change Optional to Ternary

KrisNey-MSFT:
`Change` 'optional' to 'ternary' according to last Behavioral Model meeting.
[Issue 575 Change dash_acl](https://github.com/sonic-net/DASH/issues/575)


KrisNey-MSFT:
This work is going to take some time and it also changes the attributes of the ACL table too (ternary match will add a priority attribute, which is not in design), so it will require a larger change to get it addressed.

KrisNey-MSFT:
See comments above

PR #591, opened 2024-06-14
Remove Outer SRC IP Prefix (not req's for Inbound routing lookup) in …

KrisNey-MSFT:
…sdn-features-packet-transforms.md

Per Issue 518, this is not required.  Attempting to update document.

[Issue 518 dash_pipeline.p4 Inbound Routing question](https://github.com/sonic-net/DASH/issues/518)


KrisNey-MSFT:
Inbound Routing conversations supercede this Issue

PR #594, opened 2024-06-23
[sai-gen] Maintaining the order of SAI enum types.

r12f:
Since P4 compiler won't maintain the order of the generated enums in p4runtime, it causes all enum values moves around in the code review. Since the enum types from P4 doesn't have any dependencies to each other, we can merge the new enum values into the current SAI spec by appending to the end of the list, just like other SAI attributes.

Additionally, to maintain the SAI ABI, we will not remove any existing enum values too.


mukeshmv (approved):
lgtm

PR #595, opened 2024-06-23
[ci] Update CI to ensure SAI spec are generated and updated.

r12f:
To help the code review of SAI header updates, update the CI workflow to ensure we have up-to-date SAI spec included in the PR.

This is achieved by checking git status and see if any changes in the SAI/specs folder, after running `make sai` in the CI pipeline.

### DIFF
--- a/dash-pipeline/Makefile
+++ b/dash-pipeline/Makefile
@@ -237,6 +237,14 @@ sai-clean: SAI/SAI libsai-clean saithrift-server-clean
 	rm -rf SAI/SAI/inc SAI/SAI/experimental SAI/SAI/meta
 	cd SAI/SAI && git checkout -- inc experimental meta
 
+.PHONY:check-sai-spec
+check-sai-spec:
+	if git status | grep -q "SAI/specs"; then \
+		echo 'SAI specs have changed, please run "make sai-headers" locally and commit the changes for review.'; \
+		echo ""; echo "SAI spec diff found as below:"; \
+		git diff SAI/specs; \
+		exit 1; \
+	fi
 
 # Run to recreate same environment as building saithrift client & server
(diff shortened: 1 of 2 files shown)


r12f:
Need #594 to be checked in to make it stable.

KrisNey-MSFT:
hi @chrispsommers - did you happen to check this one?  It had requested your review...TY, Kristina

dash-pipeline/Makefile
```diff
@@ -237,6 +237,14 @@ sai-clean: SAI/SAI libsai-clean saithrift-server-clean
 	rm -rf SAI/SAI/inc SAI/SAI/experimental SAI/SAI/meta
 	cd SAI/SAI && git checkout -- inc experimental meta
 
+.PHONY:check-sai-spec
+check-sai-spec:
+	if git status | grep -q "SAI/specs"; then \
```
chrispsommers, line 242: What causes the exit status to change from 1 (fail) to 0 (pass)? If the developer commits headers locally for review does this somehow remove the failure?

r12f: Hi Chris, thanks a lot for reviewing the change! 

This change is to only ensure all the SAI API changes caused by the PR will be reflected in PR as changes to SAI specs. So this task only runs in the CI pipeline and nowhere else. And the condition to pass the check is simple, as long as we have run 'make sai-headers' locally and adds the SAI spec updates in, this check will pass.

chrispsommers: ok, thanks!

chrispsommers (approved):

dash-pipeline/Makefile
r12f: Thanks Chris!

PR #596, opened 2024-06-23
[dash-p4] Deprecate order annotation on SaiVal, SaiTable and SaiCounter. (pending #594)

r12f:
Since we can now use SAI spec to handle the order of all attributes and APIs, we don't need to explicit specify the order anymore. Hence, removing it from the P4 code.

After all orders are removed, `make sai-header` locally will not cause any SAI API change as below:

```
r12f@r12f-dl380:~/code/sonic/DASH/dash-pipeline/SAI/SAI
$ gst
On branch temp
Your branch is up to date with 'origin/temp'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        bm/

nothing added to commit but untracked files present (use "git add" to track)
```


r12f:
ouch! cannot deprecate until we also have lib fixed..... closing PR.

PR #602, opened 2024-07-16
[dash-p4] Add ENI mac override support on direction lookup stage.

r12f:
This PR adds support to select which MAC address to use for looking up the ENI in the direction lookup stage.

This change follows the ExpressRoute gateway bypass HLD defined in this PR here: https://github.com/sonic-net/DASH/pull/601.


marian-pritsak (approved):

KrisNey-MSFT:
@r12f  - this has been reviewed and approved.  Would you like me to merge it for you?

r12f:
Ouch! I am seeing some conflicts on this PR now... I wonder if Marian could help me solving the conflict, so we can get it merged?

@marian-pritsak

PR #603, opened 2024-07-16
[dash-p4] Add reverse routing table and reverse tunnel support.

r12f:
The reverse routing table and reverse tunnel are added for supporting the express route gateway bypass scenario for handling the MSEE failover.

The design follows the HLD defined here: https://github.com/sonic-net/DASH/pull/601.

This PR adds the P4 code for API generation and the pipeline design. And the tests will be updated later, due to time limitation.

There are a few known issues in the current implementation:

1. Due to flow API is not fully supported in the P4, the reverse tunnel info could not be applied to the packet, as it only affects the return packets.
2. The ECMP support is not simple to support with our API design, because we don't know the count of the next hops in this tunnel in P4 pipeline. Either we need to have a predefined size, so we can create a fully populated table, which complicates the libsai generation a ton. Or, we somehow pass in the count, whenever a new member is added.

The 2 known issues are marked as TODO in the code.

### DIFF
--- a/dash-pipeline/SAI/specs/dash_eni.yaml
+++ b/dash-pipeline/SAI/specs/dash_eni.yaml
@@ -598,6 +598,19 @@ sai_apis:
     valid_only: null
     is_vlan: false
     deprecated: false
+  - !!python/object:utils.sai_spec.sai_attribute.SaiAttribute
+    name: SAI_ENI_ATTR_OUTBOUND_REVERSE_ROUTING_GROUP_ID
+    description: Action parameter outbound reverse routing group id
+    type: sai_object_id_t
+    attr_value_field: u16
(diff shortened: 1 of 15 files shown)


KrisNey-MSFT:
hi @marian-pritsak , @mukeshmv - would you be able to review? 
TY!

KrisNey-MSFT:
just checking in to see if we can have a review please?
@marian-pritsak , @mukeshmv

marian-pritsak (approved):

KrisNey-MSFT:
Looks like we still have a conflict @r12f

KrisNey-MSFT:
hi @r12f - do we need a merge on this one since the checks have passed?

KrisNey-MSFT:
Closing per Riff during Community call - switched to flow learning

PR #605, opened 2024-07-17
[doc][dash-p4] Update metering bucket design

r12f:
## Problem

The current metering design uses SAI object to identify each metering bucket. This is not efficient from both CPU and memory perspective, because of multiple reasons:

1. Whenever an ENI is created, we need to use a loop to create all metering buckets. Say, in future, if each ENI supports 1M metering buckets, we need to call the create meter bucket API 1M times.
2. The SWSS need to maintain all metering bucket object ids, which takes memory. The metering class and bucket are essentially the same or having 1:1 mapping, so this could be simplified.

To solve these issues, we are changing the metering bucket from SAI object to SAI table entry. With this, we could add assumption that all metering buckets will be created along with the ENI without additional calls, which saves all the CPU and memory resources.


r12f:
ok, looks like the SAI API doesn't have the stats API support for the table entries. I could not find any existing SAI API using this pattern. This change will need more time to change now.

image

mukeshmv:
After fixing the SAI/meta/style.pm to allow object entry stats, running into the following error.
`ERROR: SAI_OBJECT_TYPE_METER_BUCKET is not defined in OBJTOAPIMAP, missing sai_XXX_api_t declaration?
`   
We may need to push a direct commit to OCP SAI to replace the existing experimental METER_BUCKET object with the METER_BUCKET_ENTRY.

r12f:
This is because the entry stats functions are not generated in libsai. Feel free to take a look at the saiapi.cpp.j2. It should be straightforward to add (by copying the sai object one and do some small modifications)

r12f:
Better not do direct commit into SAI repo and diverge these 2 repos.

mukeshmv:
> e the entry stats functions are not generated

@r12f, No its not about the libsai entry stats function. This is OCP SAI meta error complaining about the missing SAI API headers for the original Meter bucket object type which is not removed. 
With the Dash P4 changes in this PR we are now generating the APIs only for the new Meter bucket table entry instead of the original bucket object.

mukeshmv:
> Better not do direct commit into SAI repo and diverge these 2 repos.

I was thinking to create SAI OCP PR with the new Meter bucket table entry APIs exactly as it is generated here. 
So it should not be a diverge.

r12f:
Hi Mukesh, I see. The relationship between DASH SAI and OCP SAI is that - almost at any moment, the SAI API in DASH repo is always a superset of the OCP SAI. But once we updates the API directly there, everyone in DASH repo will produce the SAI without the new meter bucket API, unless they cherry pick the changes in this PR, hence the diverge.

Since what we need to do is to pass the SAI checks, could we only submit the change that will make this PR pass to SAI, then bring the SAI back to DASH to unblock us? This is what we do for solving all previous SAI related issues too, such as spellchecks, SAI style checks and etc.

KrisNey-MSFT:
Not easy to fix.  Update OCP SAI first, move SAI submodule to the latest.  Per DASH Community call, review and accept, and push into OCP SAI.

mukeshmv:
@r12f the SAI PR has been merged. Can you update the SAI submodule to latest in this PR ?

r12f:
Hi Mukesh, ok. Do you mind to help with it? My internet access is horrible. Opening this page took me more than 10mins or so to retry... I don't feel I can get any real work done...

r12f:
I have commented the SAI PR as well. We will need get SAI merged back to DASH and get this PR merged. @mukeshmv

mukeshmv:
@r12f sure we will take care of this. As I explained before just fixing the style.pm was not sufficient to get this PR in as we were running into API missing errors. 

Since I don't have permission to add commit to this PR, I have created a duplicate PR with all your commits + updating to latest SAI. 
https://github.com/sonic-net/DASH/pull/611 - all checks have passed here. 
@prsunny merging #611 should resolve this issue.

r12f:
Perfect! Will try my best to open it and check it! Thanks Mukesh! (Do you still remember a mobile network called 3G? XD)

KrisNey-MSFT:
ROFL 😊3G

mukeshmv:
> Perfect! Will try my best to open it and check it! Thanks Mukesh! (Do you still remember a mobile network called 3G? XD)

3G - Which corner of the world are you in :)

PR #616, opened 2024-08-23
Add SAI appliance object, local region ID field

mukeshmv:
Corresponds to the local region id field added to the Appliance table schema [here](https://github.com/sonic-net/SONiC/pull/1783)

### DIFF
--- a/dash-pipeline/SAI/specs/dash_appliance.yaml
+++ b/dash-pipeline/SAI/specs/dash_appliance.yaml
@@ -1,0 +1,35 @@
+!!python/object:utils.sai_spec.sai_api_group.SaiApiGroup
+name: dash_appliance
+description: DASH appliance
+api_type: overlay
+sai_apis:
+- !!python/object:utils.sai_spec.sai_api.SaiApi
+  name: dash_appliance
+  description: DASH appliance
+  is_object: true
+  enums: []
+  structs: []
+  attributes:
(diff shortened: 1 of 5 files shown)


dash-pipeline/SAI/specs/dash_appliance.yaml
```diff
@@ -0,0 +1,35 @@
...
+sai_apis:
+- !!python/object:utils.sai_spec.sai_api.SaiApi
+  name: dash_appliance
+  description: DASH appliance
+  is_object: true
+  enums: []
+  structs: []
+  attributes:
+  - !!python/object:utils.sai_spec.sai_attribute.SaiAttribute
+    name: SAI_DASH_APPLIANCE_ATTR_LOCAL_REGION_ID
```
r12f, line 14: can we add this attributes to SAI switch instead of creating a new object?

the switch attributes are essentially for all the global attributes.

mukeshmv: Adding it to switch object may result in additional dependencies for switch object. For eg: if local region ID is immutable, then this would mean postponing SAI switch create until this value is available. Hence better to avoid this if possible.

r12f: I think immutable is not really the reason. The main reason of not putting it in the SAI switch is because this ID is coming from SDN controller, which is not available when SAI switch is created.

r12f (approved):
lgtm.

PR #617, opened 2024-08-29
[dash-flow] Rename ENI_ADDR to ENI_MAC

zhixiongniu:
The term ENI_ADDR was found to be unclear, so we have updated it to ENI_MAC in both the code and documentation.


r12f (commented):
API is not merged into SAI yet. Renaming looks good to me!

PR #619, opened 2024-09-02
[doc] Simplify MSEE failover handling in ER gateway bypass design.

r12f:
The current MSEE failover handling in ER gateway bypass design is reverse route table based, which is flexible, powerful, but a bit complicated. 

With this change, we are going to simplify the MSEE failover handling using tunnel learning. There are 1 minor issue we could not address with this design, but we will update the design later which is compatible with this design and having the problem solved.

### DIFF
--- a/documentation/express-route-service/express-route-gateway-bypass.md
+++ b/documentation/express-route-service/express-route-gateway-bypass.md
@@ -141,13 +141,16 @@ Furthermore, when MSEE failover, we need to update the reverse tunnel on all exi
 
 #### 5.2.2. MSEE device selection
 
-To handle this, we learn the PLS-to-MSEE tunnel information from the first packet when flow is created, including the encap type, key, and destination IP, and create the reverse tunnel in the reverse flow, which make sure the return traffic will be sent back to the originating MSEE.
+Unlike VM-to-* scenarios, the reverse tunnel can only have 1 single destination IP that specified using the `SAI_ENI_ATTR_VM_UNDERLAY_DIP` attribute. MSEE devices can have active-active pairs, so this single IP solution won't work.
 
-In some cases, we will need the ability to specify the source IP of the reverse tunnel. To support this, we will add a new attribute in the ENI object:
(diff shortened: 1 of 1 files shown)


marian-pritsak (changes requested):
It is not a good design to have an implicit forwarding decision. If you think that is a desired behavior, DASH should have at least a setting to enable or disable the tunnel DIP learning. Then we can support other options like ECMP and be able to choose between them.

documentation/express-route-service/express-route-gateway-bypass.md
```diff
@@ -138,92 +135,29 @@ typedef enum _sai_direction_lookup_entry_attr_t {
...
+When MSEE failover, it could impact all flows that are created. In this case, we will leverage the full flow resimulation API in the existing [flow resimulation APIs](../dataplane/dash-flow-resimulation.md) to help us keep the reverse tunnel updated.
 
-Whenever the reverse tunnel id is set, we consider the reverse tunnel routing action bit is set in the pipeline, this allows us to control the flow resimulation behavior using the same mechanism.
+Following the flow resimulation design, all flow resimulation request will be explicitly requested by the SDN controller. Hence, when MSEE failover, the SDN controller will request a full flow resimulation that marks all flows as resimulation needed.
 
-1. Step 1: Update the DIP list in the reverse tunnel object, which will cover all the new flows.
-2. Step 2: Request a full flow resimulation, which will cover the existing flows.
+When the next packet from MSEE comes in for a flow, the flow will be resimulated and the reverse tunnel will be updated accordingly, by picking up the new MSEE tunnel information.
```
marian-pritsak, line 160: What if the next packet is from the opposite side of the connection?

r12f: it is expected to forward the packet according to the non-updated flow.

let me see if this is missed in the doc.

r12f: found it! this behavior is captured in the section 5.2.3.3. below.

marian-pritsak: The packet is already in slow path because the flow it belongs to was marked for resimulation. Slow path doesn't know what the old destination was.

r12f: The old destination should be known, because it is in the existing flow.

marian-pritsak (approved):

r12f:
thanks Marian!

PR #622, opened 2024-09-28
[dash-p4] Add tunnel member table and tunnel next hop table.

r12f:
To support the ECMP group in the tunnel, this change adds the tunnel member and tunnel next hop tables using the same way as how next hop group works today in SAI.

The detailed design is following the HLD defined here: https://github.com/sonic-net/DASH/blob/main/documentation/private-link-service/private-link-service.md.


r12f:
hi @prsunny , here is the preview of the tunnel member and next hop change. please let me know, if you have any concerns. The API should be ok, and I will get the P4 part a bit more stable later.

prsunny (approved):
lgtm

mukeshmv:
@r12f , @prsunny this ECMP API design with tunnel nexthops and tunnel members being pushed iteratively is very hard to implement in the pipeline. Ideally we need to be able to pre-allocate the resources for the ECMP group.

r12f:
hi @mukeshmv , the tunnel design follows exactly the same pattern as the current SAI next hop group and next hop design. If preallocation is required, we can ensure the tunnel next hops are all created before bind the tunnel to any mappings. And after that, the next hop list will not be change. This should solve the problem.

However, we do have case there tunnel next hops being changed, say a destination being removed. Without certain support, this can fundamentally break certain scenarios. So let's touch base offline and see how to get this supported.

r12f:
Actually, after second thought, this might be related to the hash distribution thing, not the actual element count. Adding a size should help. Yea, we can add that. I will push a new PR late today.

dash-pipeline/SAI/specs/dash_tunnel.yaml
```diff
@@ -49,6 +49,32 @@ sai_apis:
...
+    is_vlan: false
+    deprecated: false
+  - !!python/object:utils.sai_spec.sai_attribute.SaiAttribute
+    name: SAI_DASH_TUNNEL_ATTR_MAX_MEMBER_SIZE
+    description: Action parameter max member size
+    type: sai_uint32_t
+    attr_value_field: u32
+    default: '1'
+    isresourcetype: false
+    flags: CREATE_AND_SET
```
mukeshmv, line 72: This should be CREATE_ONLY. Max members should not be allowed to be changed after group create.

r12f: yep and done, along with the tunnel member attribute flags.

KrisNey-MSFT:
Looks ready to merge

dash-pipeline/SAI/specs/dash_tunnel.yaml
mukeshmv: Can we please make the SAI_DASH_TUNNEL_ATTR_DASH_ENCAP and SAI_DASH_TUNNEL_ATTR_TUNNEL_KEY create-only as well ?

dash-pipeline/SAI/specs/dash_tunnel.yaml
```diff
@@ -49,6 +49,32 @@ sai_apis:
     valid_only: null
     is_vlan: false
     deprecated: false
+  - !!python/object:utils.sai_spec.sai_attribute.SaiAttribute
```
mukeshmv, line 52: Can we please make the SAI_DASH_TUNNEL_ATTR_DASH_ENCAP and SAI_DASH_TUNNEL_ATTR_TUNNEL_KEY create-only as well ?

prsunny: @r12f , can you check?

prsunny:
> @prsunny Can you please specify the scale parameters for these objects in the HLD number of DASH_TUNNEL objects, number of members per DASH_TUNNEL object.

@mukeshmv , the following are the scale we are thinking:
1. DASH_TUNNEL - 1k per dpu
2. MEMBER scale - 128 

Will update the sonic-hld for this scale numbers

dash-pipeline/SAI/specs/dash_tunnel.yaml
KrisNey-MSFT: Tunnel s/be updatable?  Unlikely that we will update a VNI inside of a tunnel, however IPs inside may change (create-set).  (Key=VNI).
Would we have a case where we change the key (MSEE for example)?  Yes, we have seen this before.  Can be handled as tunnel replacement on the route.  

@r12f and @mzms to finalize - s/be ok to be create-only

prsunny: I think this can be create_only. I don't see a case where we want to update VNI on the fly. If its required, we can create new tunnel with new members, bind and then remove the old tunnel.

mukeshmv: Yes, since the DASH TUNNEL is now a group of tunnels, modifying encap/VNI on all tunnels in the group is non-trivial.

IP change should be fine since my understanding is that any IP change in the northbound DASH_TUNNEL will convert to delete old DASH_TUNNEL_MEMBER and then add new DASH_TUNNEL_MEMBER at the SAI API level. This is fine as long as the total members stays within the max member size configured initially.

r12f: Updated now.

mukeshmv (approved):
lgtm

PR #630, opened 2024-10-12
[debian]: Add saimetadata to libsai debian package

Pterosaur:
Add shared library saimetadata to libsai debian package. Because when a program is linked to a sai library(such as dash sai), it should also be linked to the corresponding saimetadata library as well. Otherwise, there may be a risk of crashing if the sai library uses an attribute that is not in the saimetadata library.


r12f:
hi @Pterosaur , do you mind to add a description on why doing this change?

Pterosaur:
@r12f , please check it again

r12f (approved):

PR #631, opened 2024-10-15
Add HA session parameter Bounceback IP

marian-pritsak:
Add bounceback IP which is a different IP from
peer IP, to emulate the same network path for data packets as for the SYN packets.

### DIFF
--- a/dash-pipeline/SAI/specs/dash_ha.yaml
+++ b/dash-pipeline/SAI/specs/dash_ha.yaml
@@ -140,6 +140,19 @@ sai_apis:
     valid_only: null
     is_vlan: false
     deprecated: false
+  - !!python/object:utils.sai_spec.sai_attribute.SaiAttribute
+    name: SAI_HA_SET_ATTR_PEER_BOUNCEBACK_IP
+    description: Action parameter peer bounceback IP
+    type: sai_ip_address_t
+    attr_value_field: ipaddr
+    default: 0.0.0.0
+    isresourcetype: false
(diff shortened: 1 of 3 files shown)


marian-pritsak:
Depends on https://github.com/opencomputeproject/SAI/pull/2090

KrisNey-MSFT:
hi @marian-pritsak - is an HLD update in play for this week?

KrisNey-MSFT:
@marian-pritsak  https://github.com/opencomputeproject/SAI/pull/2090 has been merged - is it ok to merge this one too?

r12f:
Hi Marian, do you mind explaining a bit more on what does this attribute do? @KrisNey-MSFT , maybe we can do it in one of our community call. 

Junhua is helping bring up the SAI API in DASH to latest, but this difference is causing problems, since the attribute is not aware in DASH, do generated SAI will simply revert this.

KrisNey-MSFT:
Do not need to send HA back to DPU; this is an optimization.  To preserve entropy and look as if it is coming from the same switch.  @marian-pritsak will update the HA document in the SONiC repo.  Attribute to support capability queries.  If attribute not supported, then not enforced.  (SAI API is already merged).

KrisNey-MSFT:
Hi @marian-pritsak , has the HA document been updated in the SONiC repo?  @r12f , and/or @mukeshmv as Reviewers

KrisNey-MSFT:
Hi @r12f  and @mukeshmv  - is this ready to review?

PR #632, opened 2024-10-20
[libsai] Make helper function for bulk object create and removal to simplify libsai generation.

r12f:
All bulk create/removal objects SAI implementation is essentially the same. Hence making some helper functions for simplify the generated libsai.

With more changes coming, we should be able to make the libsai templates and generations more maintainable.


jimmyzhai (approved):

r12f:
thanks Junhua!

PR #638, opened 2024-11-01
Add an internal config table to apply unexposed configurations to switch before testing

ShaofengWu123:
This PR renamed the table `underlay_mac` to `internal_config`. The new `internal_config` table is ignored for SAI API generation and will be used for setting switch to a initial state required by some test cases.
For example, setting `flow_enabled` to `False` will cause packet to bypass the flow table in the pipeline. In order to test dash flow, we can set `flow_enabled` to `True` before corresponding tests, e.g. ptf test `saidashflow.py`.


dash-pipeline/SAI/src/dashsai.cpp
```diff
@@ -176,6 +186,69 @@ sai_status_t DashSai::apiInitialize(
...
+
+    if (grpc::StatusCode::OK != retCode)
+    {
+        DASH_LOG_ERROR("Failed to insert entry into internal_config table");
+        return SAI_STATUS_FAILURE;
+    }
+
+    printf("Inserted entry into internal_config table successfully");
+
     return SAI_STATUS_SUCCESS;
```
jimmyzhai, line 252: Better to add individual test fixture to update internal_config, instead of in apiInitialize

jimmyzhai (approved):

PR #642, opened 2024-11-08
Update the scale numbers for ExpressRoute gateway bypass scenario.

r12f:


documentation/express-route-service/express-route-gateway-bypass.md
```diff
@@ -77,13 +77,12 @@ The scaling requirement for PL redirect map are listed as below. The metrics are
 
 | Metric | Requirement |
 | ------ | ----------- |
-| # of ENIs per DPU | 32 |
+| # of ENIs per DPU (every 200G) | 32 |
 | # of VNET mapping per ENI | 64K |
 | # of PPS | 64M |
-| VNET mapping change rate (CRUD) | (TBD) |
+| # of VNI per ENI | 1 |
+| # of MSEE source PA per ENI | 64 |
```
kperumalbfn: Could you also update number of MSEEs per region(64).

kperumalbfn (approved):

documentation/express-route-service/express-route-gateway-bypass.md
r12f: discussed offline and to avoid making it more complicated, as region changes over time, we put the final number in the requirement, which takes region already into considerations.

r12f:
thanks Kumaresh!

PR #645, opened 2024-11-12
clarification for Calculation error in program scale testing requirements?

mgheorghe:
clarification for https://github.com/sonic-net/DASH/issues/397


KrisNey-MSFT:
hi @mgheorghe  - can we close 397 then?

mgheorghe:
Yes, let's close it; unfortunately, I don't have the right to do it myself.

PR #647, opened 2024-12-05
[sai-gen] Update SAI specs to have more p4 meta info

jimmyzhai:
This PR updates p4 meta of sai api to include table keys and action params in detail.
It will be the single truth to generate sai api implementation later.


r12f (approved):
looks straight forward enough and does contains a good amount of the info that we need to generate the dash libsai. thanks Junhua!

Issue #436, opened 2023-09-09
Why dash_sai_create_dash_acl_rule create multiple SAI_OBJECT_TYPE_DASH_ACL_RULE objects?


kcudnik:
When sai implementation is generated, functions is added:
```c
 176 static sai_status_t dash_sai_create_dash_acl_rule(
 177         _Out_ sai_object_id_t *dash_acl_rule_id,
 178         _In_ sai_object_id_t switch_id,
 179         _In_ uint32_t attr_count,
 180         _In_ const sai_attribute_t *attr_list)
```
in generated file: dash-pipeline/SAI/lib/saidashacl.cpp

but internally i see 5 times, so 5 times object is created;
```c
objId = dashSai->getNextObjectId((sai_object_type_t)SAI_OBJECT_TYPE_DASH_ACL_RULE);
```

and then inserted:
```c
if (false == dashSai->insertInTable(matchActionEntry, objId)) {
```

but insertion is with the 5 different objects,

that generated code for that functions process in loop several attributes which are duplicated in each process line

can anyone explain my why this is happening ? this function should create only 1 single SAI objects, perhaps multiple bmv2 objects internally, but from SAI perspective only one,  i see this as a bug, **please explain**

chrispsommers:
@marian-pritsak can best explain this, it's a byproduct of some implementation choices, not a "bug."

kcudnik:
further more if we grep actionId:
```c
  24     pi_p4_id_t actionId = 0;
  45     actionId = 25655048; // SAI_DASH_ACL_GROUP_ACTION_SET_ACL_GROUP_ATTRS
  48     action->set_action_id(actionId);
 193     pi_p4_id_t actionId = 0;
 283                     actionId = 32161567;
 289                     actionId = 20706700;
 295                     actionId = 28146588;
 301                     actionId = 31424218;
 314     action->set_action_id(actionId);
 423                     actionId = 32161567;
 429                     actionId = 20706700;
 435                     actionId = 28146588;
 441                     actionId = 31424218;
 454     action->set_action_id(actionId);
 563                     actionId = 32161567;
 569                     actionId = 20706700;
 575                     actionId = 28146588;
 581                     actionId = 31424218;
 594     action->set_action_id(actionId);
 703                     actionId = 18858683;
 709                     actionId = 24263137;
 715                     actionId = 29962337;
 721                     actionId = 26077229;
 734     action->set_action_id(actionId);
 843                     actionId = 18858683;
[...]

PR #653, opened 2024-12-19
[libsai] Re-implement APIs create/remove

jimmyzhai:
Following https://github.com/sonic-net/DASH/pull/651, refactor APIs create/remove for easier read and maintenance:

1. Use common code `DashSai::create` and `DashSai::remove` to implement them for all SAI objects/entries
2. Fix issue https://github.com/sonic-net/DASH/issues/436, now only 1 SAI object is created for multiple bmv2 objects (each bmv2 object as a p4 table entry in p4 table for each stage). 
3. Fix issue https://github.com/sonic-net/DASH/issues/654

The generated sample code of APIs create/remove attribute is as below:

- SAI object objectID
image

- SAI object entry
image

For the case of acl rule object, one acl rule is inserted into multiple p4 tables in different stages. These p4 stage tables are same and the only difference is `table id` and `table action id` in table entry, captured in struct P4MetaSiblingTable:

```
    struct P4MetaSiblingTable
    {
        uint32_t id;
        // action enum id -> p4 action id
        std::map<uint32_t, uint32_t> actions;
    };
```
The generated code of sibling tables for acl rule is as below:
image


dash-pipeline/SAI/src/p4meta.cpp
```diff
@@ -0,0 +1,407 @@
...
+
+    const P4MetaKey* P4MetaTable::get_meta_key(
+            _In_ const std::string &key_name) const
+    {
+        for (auto i=0u; i<keys.size(); i++) {
+            if (key_name == keys[i].name) {
+                return &keys[i];
+            }
+        }
+
```
kcudnik, line 33: over entire PR incosistent placemtn { once in end of function once in new line, over all project has { in new line

dash-pipeline/SAI/src/p4meta.cpp
```diff
@@ -0,0 +1,407 @@
...
+        else if (key.match_type == "list") {
+            auto mf_optional = mf->mutable_optional();
+            get_attr_value_from_p4(key.field, key.bitwidth, mf_optional, value);
+        }
+        else if (key.match_type == "range_list") {
+            auto mf_optional = mf->mutable_optional();
+            get_attr_value_from_p4(key.field, key.bitwidth, mf_optional, value);
+        }
+        else {
+            assert(0);
```
kcudnik, line 283: some error message would be nice

dash-pipeline/SAI/src/p4meta.h
```diff
@@ -0,0 +1,256 @@
...
+           int32_t val = *(const int32_t*)v;
+           value.s32 = ntohl(val) >> (32 - bitwidth);
+        }
+        else if (field == "u32") {
+           uint32_t val = *(const uint32_t*)v;
+           value.u32 = ntohl(val) >> (32 - bitwidth);
+        }
+        else if (field == "u64") {
+           uint64_t val = *(const uint64_t*)v;
+           if (*reinterpret_cast<const char*>("\0\x01") == 0) { // Little Endian
```
kcudnik, line 148: what is that ?

dash-pipeline/SAI/src/p4meta.cpp
jimmyzhai: fixed

dash-pipeline/SAI/src/p4meta.cpp
jimmyzhai: fixed

dash-pipeline/SAI/src/p4meta.h
jimmyzhai: A quick approach to checking host arch endianness, from https://github.com/sonic-net/DASH/blob/main/dash-pipeline/SAI/src/utils.h#L143.

kcudnik: please don't use that, use proper macros/functions defined endian.h? and make proper function for that if you need running time host endian

r12f (approved):
this change looks good to me! do you mind to add an example of the generated sibling table related metadata into the PR description? It will be great for people to understand the change.

dash-pipeline/SAI/src/dashsai.cpp
```diff
@@ -1043,6 +1049,169 @@ sai_status_t DashSai::bulk_remove_objects(
...
+        DASH_LOG_ERROR("Not find p4 table action");
+        return SAI_STATUS_FAILURE;
+    }
+
+    auto action = matchActionEntry->mutable_action()->mutable_action();
+    action->set_action_id(action_id);
+
+    for (uint32_t i = 0; i < attr_count; i++)
+    {
+        if (auto meta_param = meta_table.get_meta_action_param(action_id, attr_list[i].id))
```
r12f, line 1099: will be better to move the assignment out of the if. this code pattern is easy to go wrong in general.

dash-pipeline/SAI/src/dashsai.cpp
```diff
@@ -1194,29 +1395,63 @@ sai_status_t DashSai::get(
...
+        _In_ p4::v1::Update_Type updateType,
+        _In_ uint32_t action_id)
+{
+    if (meta_table.sibling_tables.empty())
+    {
+        return;
+    }
+
+    std::shared_ptr<p4::v1::TableEntry> entry = std::make_shared<p4::v1::TableEntry>();
+    entry->CopyFrom(*matchActionEntry);
```
r12f, line 1441: will be better to add a comment here to indicate all key and attribute index/id are the same.

r12f: only the action is different.

jimmyzhai: Added comments

dash-pipeline/SAI/src/dashsai.cpp
jimmyzhai: fixed

jimmyzhai:
> this change looks good to me! do you mind to add an example of the generated sibling table related metadata into the PR description? It will be great for people to understand the change.

Updated PR description to have the generated sibling tables for acl rule.

kcudnik (approved):

jimmyzhai:
fixed issue #436

PR #659, opened 2025-01-14
Add FNIC pics

KrisNey-MSFT:
Adding two picture files for DASH in VM mode, and DASH in FNIC mode


KrisNey-MSFT:
Per Community call, just merging a couple of pictures

PR #660, opened 2025-01-20
Add tunnel type control between VM and SmartSwitch

marian-pritsak:
Allow choosing between different encapsulation
types for the tunnel used in inbound direction.

### DIFF
--- a/dash-pipeline/SAI/specs/dash_eni.yaml
+++ b/dash-pipeline/SAI/specs/dash_eni.yaml
@@ -639,6 +639,19 @@ sai_apis:
     valid_only: null
     is_vlan: false
     deprecated: false
+  - !!python/object:utils.sai_spec.sai_attribute.SaiAttribute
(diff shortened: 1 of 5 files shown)


dash-pipeline/SAI/specs/dash_eni.yaml
```diff
@@ -639,6 +639,19 @@ sai_apis:
     valid_only: null
     is_vlan: false
     deprecated: false
+  - !!python/object:utils.sai_spec.sai_attribute.SaiAttribute
+    name: SAI_ENI_ATTR_INBOUND_DASH_ENCAPSULATION
```
r12f, line 643: Hi Marian, may I know which scenario is this? In my understanding, we need to support both encapsulations in the inbound side at the same time, so this solution might not work.

KrisNey-MSFT:
hi @marian-pritsak - per @r12f  question above...

Hi Marian, may I know which scenario is this? In my understanding, we need to support both encapsulations in the inbound side at the same time, so this solution might not work.

PR #661, opened 2025-01-21
Override VM MAC when in DST address ENI lookup

marian-pritsak:


KrisNey-MSFT:
Result of testing at NVidia.  Needs to also set SRC Overlay MAC..?  Possibly... Marian to check and update PR.

KrisNey-MSFT:
hi @marian-pritsak - do we still need this one?

KrisNey-MSFT:
Closing as no longer relevant, per @marian-pritsak

PR #669, opened 2025-02-25
[dash-p4] Add flow_table_id attribute on ENI

jimmyzhai:
By adding `flow_table_id` attribute on ENI object, ENI can be associated with a specific flow table

### DIFF
--- a/dash-pipeline/SAI/specs/dash_eni.yaml
+++ b/dash-pipeline/SAI/specs/dash_eni.yaml
@@ -639,6 +639,19 @@ sai_apis:
     valid_only: null
     is_vlan: false
     deprecated: false
+  - !!python/object:utils.sai_spec.sai_attribute.SaiAttribute
+    name: SAI_ENI_ATTR_FLOW_TABLE_ID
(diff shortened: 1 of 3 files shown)


dash-pipeline/bmv2/dash_headers.p4
```diff
@@ -241,6 +241,7 @@ header flow_data_t {
     bit<32> version;
     dash_flow_action_t actions;
     dash_meter_class_t meter_class;
+    bit<32> idle_timeout;
```
r12f, line 244: will be better to split this PR into 2 and only keep the flow table id in this PR.

jimmyzhai: Done

jimmyzhai: Another one: https://github.com/sonic-net/DASH/pull/670

r12f (approved):
lgtm also confirmed with Marian and Mukesh offline, both looking good! thanks Junhua!

PR #670, opened 2025-02-26
[dash-p4] Add flow idle timeout support in behavior model

jimmyzhai:
Support customized flow idle timeout in flow creation. The value can be propagated from `flow_ttl_in_milliseconds` attribute of the flow table. Also update flow entry key selection from `flow_enabled_key` attribute of the flow table.

### DIFF
--- a/dash-pipeline/bmv2/dash_headers.p4
+++ b/dash-pipeline/bmv2/dash_headers.p4
@@ -241,7 +241,7 @@ header flow_data_t {
     bit<32> version;
     dash_flow_action_t actions;
     dash_meter_class_t meter_class;
-    bit<32> idle_timeout;
+    bit<32> idle_timeout_in_ms;
 }
 const bit<16> FLOW_DATA_HDR_SIZE=flow_data_t.minSizeInBytes();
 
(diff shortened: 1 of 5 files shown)


dash-pipeline/bmv2/dash_headers.p4
```diff
@@ -241,6 +241,7 @@ header flow_data_t {
     bit<32> version;
     dash_flow_action_t actions;
     dash_meter_class_t meter_class;
+    bit<32> idle_timeout;
```
r12f, line 244: _in_ms

dash-pipeline/dpapp/dash/flow.h
```diff
@@ -11,6 +11,8 @@
 
 /* Default timeout in seconds */
 #define DASH_FLOW_TIMEOUT   30
+/* Max timeout in seconds */
+#define DASH_FLOW_MAX_TIMEOUT   1024
```
r12f, line 15: 1800 or 2048, both are good.

jimmyzhai: Updated it to 1800

dash-pipeline/bmv2/dash_headers.p4
jimmyzhai: fixed

r12f (approved):
thanks a lot Junhua! lgtm!

PR #671, opened 2025-03-14
Add single_match_priority in SaiTable annotation

jimmyzhai:
If at least one of the match fields is Optional, Ternary or Range, p4runtime v1 model requires 'priority' to be set for each table entry. But in some dash scenarios, 'priority' is not required and is not exposed to dash sai. For these special cases, we add a single_match_priority field in SaiTable annotation to mark it. If it is 'true', the behavior changes are: 

- Not generate attribute PRIORITY in SAI object
- Not generate field priority in SAI entry struct
- Set the priority of all table entries to a default (fixed) value 1 in dash p4 libsai

It also adds the set/get support of p4 match kind `range`.

### DIFF
--- a/dash-pipeline/SAI/utils/dash_p4/dash_p4_table.py
+++ b/dash-pipeline/SAI/utils/dash_p4/dash_p4_table.py
@@ -33,6 +33,7 @@ class DashP4Table(DashP4Object):
         self.stage: Optional[str] = None
         self.is_object: Optional[str] = None
         self.api_type: Optional[str] = None
+        self.single_match_priority: Optional[bool] = False
 
     def parse_p4rt(
         self,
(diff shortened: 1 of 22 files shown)


dash-pipeline/SAI/utils/dash_p4/dash_p4_table.py
```diff
@@ -419,19 +413,18 @@ def create_sai_attributes(self, sai_api: SaiApi) -> None:
...
-            if any([key.match_type != "exact" for key in self.keys]) and all(
-                [key.match_type != "lpm" for key in self.keys]
-            ):
-                priority_attr = SaiAttribute(
-                    name=f"SAI_{self.name.upper()}_ATTR_PRIORITY",
-                    description="Rule priority in table",
-                    type="sai_uint32_t",
-                    flags="MANDATORY_ON_CREATE | CREATE_ONLY",
-                )
+        # If any match key in this table supports priority, we need to add a priority attribute.
```
r12f, line 416: this code only covers the attribute part, but not the structure part, hence it results in existing priority entry being removed.

jimmyzhai: Updated. Also cover structure part.

PR #698, opened 2026-04-28
add eni attribute to get ha flow owner (DPU-driven HA)

zjswhhh:
Today for DPU-driven HA, there is no way to tell which side is the actual operational owner of HA, both sides will show as HA state active. Adding an attribute to get the flow owner, aka. the actual active. 

SAI: https://github.com/opencomputeproject/SAI/pull/2278

sign-off: Jing Zhang zhangjing@microsoft.com


documentation/high-avail/ha-api-hld.md
```diff
@@ -226,6 +226,7 @@ To provide the ENI-level HA control, each ENI will have the following SAI attrib
 | -------------- | ---- | ----------- |
 | SAI_ENI_ATTR_HA_SCOPE_ID | `sai_object_id_t` | The HA scope ID of the ENI. |
 | SAI_ENI_ATTR_IS_HA_FLOW_OWNER | `bool` | Determines which DPU in the pair creates flows belonging to this ENI in steady-state. Typically this is set to True for the ENIs on the Active DPU and False for the Standby DPU. |
+| SAI_ENI_ATTR_IS_OPERATIONAL_HA_FLOW_OWNER | `bool` | Read-only. Returns the operational HA flow ownership status of the ENI. |
```
yue-fred-gao, line 229: This is at eni level. How are you planning to report ha-state which is at dpu level?

zjswhhh: Swss can aggregate? Do we want the raw sai attribute values written to state db for troubleshooting purpose?

